PR #11, opened 2015-07-30
Doc updates for 1.4.

garyo:
I re-enabled some #ifdefs but only when DOXYGEN is defined, and set
doxygen up to define that symbol.  This way the deprecated/removed
functions and macros still get documented (as deprecated).

Also beefed up the genRelease script to make it easier to prep
releases and do test doc builds.

(This also removes a bunch of excess end-of-line whitespace in the files I touched.)


garyo:
This is outdated.  We decided to add ofxOld.h for old symbols.

PR #14, opened 2015-08-03
Added ofxOld.h for deprecated symbols

PierreJasmin:
Did that work?

### DIFF
--- a/include/ofxOld.h
+++ b/include/ofxOld.h
@@ -34,8 +34,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  
 /** @brief String to label images with YUVA components
 --ofxImageEffects.h
-Note, this has been deprecated in v1.3
-removed in v1.4
+@deprecated - removed in v1.4. Note, this has been deprecated in v1.3
+
  */
 #define kOfxImageComponentYUVA "OfxImageComponentYUVA"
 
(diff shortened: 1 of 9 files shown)


include/ofxOld.h
```diff
@@ -0,0 +1,99 @@
...
+#define kOfxImageComponentYUVA "OfxImageComponentYUVA"
+
+/** @brief Indicates whether an effect is performing an analysis pass.
+--ofxImageEffects.h
+   - Type - int X 1
+   - Property Set -  plugin instance (read/write)
+   - Default - to 0
+   - Valid Values - This must be one of 0 or 1
+
+This feature has been deprecated - officially commented out v1.4.
```
garyo: Could you add the @deprecated doxygen tag to these?  See my version at https://github.com/garyo/openfx/commit/8e715b2834570f9a8089f032cb8705694a9e0661?w=1 for reference.

PierreJasmin: Do we still need #ifdef DOXYGEN  ?

garyo: No.  That was only when trying to keep the #defines inline -- that's what prompted me to ask for alternative ideas.  I think all you need is the @deprecated. (Plus switch the 1.3 to 1.4, but that should be a separate commit and PR, I can do that if desired.)

PierreJasmin: Added @deprecated -  and leaving until wed., will only have email late 
at night

PierreJasmin:
Great

Will look at it this week-end
I pre-wrote an announcement, will send for review

Etc

-- Confused how this gets updated? -- 
http://openeffects.org/documentation/api_doc/
OFX API v. 1.3 Reference 
<http://openeffects.org/documentation/api_doc>   -- Useful links on 
front page

Pierre

PR #19, opened 2015-08-26
fix various issues in ofx includes

devernay:
- constify parameters and return values when appropriate
- add half sample format (required for OpenGL)
- kOfxParamPropInteractMinimumSize is a double x 2 in Nuke

### DIFF
--- a/include/ofxPixels.h
+++ b/include/ofxPixels.h
@@ -78,7 +78,7 @@ typedef struct OfxRGBColourF {
 
 /** @brief Defines a double precision floating point component RGB pixel */
 typedef struct OfxRGBColourD {
-  double r, g, b, a;
+  double r, g, b;
 }OfxRGBColourD;
 
 #ifdef __cplusplus
--- a/include/ofxCore.h
+++ b/include/ofxCore.h
@@ -85,7 +85,7 @@ typedef struct OfxHost {
          - NULL if the API is unknown (either the api or the version requested),
 	 - pointer to the relevant API if it was found
   */
-  void *(*fetchSuite)(OfxPropertySetHandle host, const char *suiteName, int suiteVersion);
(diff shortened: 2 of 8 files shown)


include/ofxPixels.h
```diff
@@ -78,7 +78,7 @@ typedef struct OfxRGBColourF {
 
 /** @brief Defines a double precision floating point component RGB pixel */
 typedef struct OfxRGBColourD {
-  double r, g, b, a;
+  double r, g, b;
```
garyo: This seems like a breaking change, since it changes the size of the struct.  I know it's supposed to be ignored, and I don't think any API function currently uses it directly (though the Support lib does), but I don't think we can just change this.

garyo:
Before merging, I think this needs another reviewer's eyes.

include/ofxPixels.h
devernay: Note that this struct is not used anywhere AFAIK, it seems to be here for convenience.
It was obviously a typo in the original version of the include (OfxRGBColourB OfxRGBColourS OfxRGBColourF were OK, as were all the RGBA versions). I noticed that when I wrote a function that actually took an RGB triplet as argument.

devernay: oh yes this is used as the return value of getBackgroundColour in the Support lib. Anyway, the support lib also has plenty of other bugs (I'll make another pull request). Changing the header should cause the lib to recompile, and it is statically linked with plugins, so everything should be OK. This struct never goes through the OpenFX API.

devernay: actually, libSupport had its own definition in ofxsCore.h:
```
Should migrate this to the ofxCore.h in a v1.1
*/
struct OfxRGBColourD {
  double r, g, b;
};
```
I'm preparing a pull request for Support, which includes fixing that too

PR #22, opened 2015-08-27
fix HostSupport

devernay:
- add OpenGL render support (via OFX_SUPPORTS_OPENGLRENDER)
- add support for a host implementation of the multithread suite (via OFX_SUPPORTS_MULTITHREAD)
- use const where appropriate, pass "const string&" instead of "string"
- add FreeBSD support
- add missing includes for portability
- add doc
- plugin identifiers are not case-insensitive
- catch plugin C++ exceptions that would accidentally pass through the (C) entry points
- many more bug and feature fixes (see log for each commit)


garyo:
I don't use HostSupport so I'm not the best one to comment on this.  Anyone?

PR #32, opened 2015-09-08
OpenGL fixes

barretpj:
Patches to fix issues with OpenGL support as requested from ofxa-members


devernay:
thanks! I missed these.

HostSupport/src/ofxhImageEffect.cpp
```diff
@@ -341,6 +342,7 @@ namespace OFX {
         { kOfxImageEffectPropSupportsTiles,     Property::eInt,        1, false, "1" }, // OFX 1.4
 #ifdef OFX_SUPPORTS_OPENGLRENDER
         { kOfxImageEffectPropOpenGLRenderSupported, Property::eString, 1, false, "false"}, // OFX 1.4
+        { kOfxOpenGLPropPixelDepth, Property::eString,  0, false, "" },
```
devernay: this should only be in the descriptor, not in the instance (just like kOfxImageEffectPropSupportedPixelDepths)

garyo:
Hi folks; I'm just cleaning up some old PRs. Is this one still valid? Happy to merge it if so.

devernay:
Still valid, except for that line which should be removed (see my comment).
Here's the trivial fix: https://github.com/NatronGitHub/openfx/commit/c82281371d15ec093d1edeaf5a16af8c5e9da684

garyo:
Hi folks; I'm not sure about the last two commits from @barretpj  -- they seem to revert the whole PR. @barretpj perhaps you intended to base this PR on a local branch, but then added extra commits which became part of the PR? 
I'm pretty sure I can figure out the intent, but would like your confirmation.
Should we keep the three commits ending at d3849c281b and then remove the one line @devernay mentioned?

barretpj:
This was a while ago, and I have no more clue about PRs now than I did then! Your suggestion sounds correct

PR #39, opened 2016-06-30
Uninitialized vars in examples

chrisdolan:
Fix several uninitialized variables and unchecked null pointers in the example utilities


garyo (approved):
LGTM (sorry for the extremely long wait!)

PR #41, opened 2016-07-18
Add nuke extension

gregoire-dl:
- Add layout options
- Option to define group as tab


fxtech:
Looks mostly low-hanging fruit. Happy to implement those that I can.
Do we take this header as-is or fold them into some other existing file(s)?

PR #54, opened 2019-10-06
MultiTread V2 suite

revisionfx:
Added V2 proposal
 
## Description

Adds OfxMultiThreadSuiteV2 which simplifies OfxMultiThreadSuiteV1 by removing the five *mutex calls which should really be done using OS calls.  The *multiThread function pointers remain as is. This is part of API clean-up to limit suite ambiguities with regards to what an host supports or not.

This is in response to feature request: "we could have a rev of the threading API, which is slightly tidier than the current one, which allows access to that thread pool. The host still needs to tell the plugin, on a per action basis, how many threads it is allowed to use."

**ALTERNATIVE:**  The other way to do this would be a new property on the inArgs, kOfxPropNumberThreadsAllowed, which would be the total number of threads the plugin could launch. This would need to be added to the InArgs of each action.  This is not a mutually exclusive proposition. This pull request only addresses OfxMultiThreadSuite.

### DIFF
--- a/include/ofxMultiThread.h
+++ b/include/ofxMultiThread.h
@@ -4,7 +4,7 @@
 /*
 Software License :
 
-Copyright (c) 2003-2009, The Open Effects Association Ltd. All rights reserved.
+Copyright (c) 2003-2019, The Open Effects Association Ltd. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


garyo:
Closed; Pierre will resubmit on branch.

PR #57, opened 2019-10-09
Feature/desired premult

revisionfx:
Currently we have the  kOfxImageEffectPropPreMultiplication(a Clip Preferences Action): kOfxImageOpaque, kOfxImagePreMultiplied, kOfxImageUnPreMultiplied
Request is to add additional effect instance clarity for cases where main input clip alpha is kOfxImageOpaque but output is not (e.g. a keyer). We realize some hosts might allow a branch of their processing graph to switch but this is just to set an host level policy that would make most sense to their users or internal processing - typically an host compositing operator is in one state or other by default.

kOfxImageEffectHostDesiredPremult (value can be kOfxImagePreMultiplied or kOfxImageUnPreMultiplied).

### DIFF
--- a/include/ofxImageEffect.h
+++ b/include/ofxImageEffect.h
@@ -992,6 +992,21 @@ See the documentation on clip preferences for more details on how this is used w
 /** Used to flag an image as unpremultiplied */
 #define kOfxImageUnPreMultiplied "OfxImageAlphaUnPremultiplied"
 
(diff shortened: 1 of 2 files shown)


revisionfx:
This should be closed - has been addressed

garyo:
Closing per Pierre's note above.

PR #59, opened 2019-10-20
Features/MultiThread_suite_V2

revisionfx:
Update the MultiThread suite to V2. Request is to clean up suites that have ambiguities in terms of what is implemented. All mutext calls in V2 are removed. Plug-in would need to use OS calls for that if needed.

## Description
Plug-in can check if V2 is supported by host, if not revert to V1 or ignore.
This suite allows hosts to manage a thread pool and fight back against over-thread subscription on multi-core systems.

### DIFF
--- a/include/ofxMultiThread.h
+++ b/include/ofxMultiThread.h
@@ -4,7 +4,7 @@
 /*
 Software License :
 
-Copyright (c) 2003-2009, The Open Effects Association Ltd. All rights reserved.
+Copyright (c) 2003-2019, The Open Effects Association Ltd. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
@@ -31,6 +31,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 
+
 #include "ofxCore.h"
 
 #ifdef __cplusplus
@@ -44,9 +45,6 @@ extern "C" {
 
 #define kOfxMultiThreadSuite "OfxMultiThreadSuite"
 
-/** @brief Mutex blind data handle
- */
-typedef struct OfxMutex *OfxMutexHandle;
 
 /** @brief The function type to passed to the multi threading routines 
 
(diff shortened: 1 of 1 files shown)


revisionfx:
There is only a single file, initially copied in wrong folder and deleted

garyo:
Just reviewing old PRs... I see the version of `include/ofxMultiThread.h` here has diff markers in it so it's clearly incorrect. If you'd like to send a proper version of what your intent is (just a clean updated version of that file for instance), I can create a new PR or update this one.

revisionfx:
Not sure what you mean by different marker. You mean when you switched from Features to Feature/ :)
It was simply suggested that Mutex should not be handled by suite.
Right now we use that suite or not on a per host basis but don't use Mutex wrappers.  Haven't scanned host by host to see what they respond to these days... 

A bigger issue for us is with hosts that can render multiple frames at once - we have kOfxImageEffectPluginRenderThreadSafety for that. For example host might not support the OpenGL suite but we can still render on GPU from CPU images, however we had issues in the past with nVidia driver in such case in some host spawning a bunch of multiple frames render at once (making this real slow - the warp engine loops and can't find a slot) so we are not then [kOfxImageEffectRenderFullySafe] - we are though if a user sets our plugin to render on CPU... Definition is only a Property Set - plugin descriptor (read/write) - does not say if one can change in Instance Changed...  and have not tested.  Alternatively in the GPU additional WIP features it was suggested we have a parallel interface for I use GPU and which one.

(https://openfx.readthedocs.io/en/latest/Reference/ofxRendering.html#ofxImageEffect_8h_1a9dd536b70cb80adac6bf6e67cb9e50b0) 

https://openfx.readthedocs.io/en/latest/Reference/DoxygenIndex.html#c.kOfxImageEffectPluginRenderThreadSafety does support

garyo:
No, I mean the file itself has "diff markers" in it. It would not compile as is. You replaced the actual file with something corrupted. The first line of include/ofxMultiThread.h is now this literal text:
```
<<<<<<< HEAD
```

garyo:
As per your other points, I suggest you take that up with a host vendor who is willing to implement their side of it, and see where that goes. I wouldn't post it in a PR until you have that.

I couldn't read the changes in this PR because of the file corruption, so I don't really know what it contains. Maybe if you can send me an un-corrupted version of what you intend the header file to look like, I can clean up this branch with it.

PR #60, opened 2019-10-20
update to ofxParam.h -  ParamTypeClip

revisionfx:
The purpose of this parameter is to logically locate in the effects controls the clip/layer param indexer and is particularly useful for hosts that don't have other means to connect additional input clips to effects. The clip creation per se remains the same and the param name should match clip name.


garyo:
Closing as duplicate of #85 .

PR #62, opened 2019-10-20
ofxImageEffect.h - Desired Premult

revisionfx:
This Property defines an host level policy. So effects receiving an ImageOpaque and Generator know to be Premult or Straight if they don't generate a full Alpha. See Standard Change issue #66.

### DIFF
--- a/include/ofxImageEffect.h
+++ b/include/ofxImageEffect.h
@@ -4,7 +4,7 @@
 /*
 Software License :
 
-Copyright (c) 2003-2019, The Open Effects Association Ltd. All rights reserved.
+Copyright (c) 2003-2020, The Open Effects Association Ltd. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
@@ -980,6 +980,7 @@ This is the actual value of the component depth, before any mapping by clip pref
       - kOfxImageUnPreMultiplied - the image is unpremultiplied
 
 See the documentation on clip preferences for more details on how this is used with the ::kOfxImageEffectActionGetClipPreferences action.
+version 1.5 and over:  See kOfxImageEffectHostDesiredPremult and how it interacts with this property.
 */
 #define kOfxImageEffectPropPreMultiplication "OfxImageEffectPropPreMultiplication"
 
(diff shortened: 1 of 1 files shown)


include/ofxImageEffect.h
```diff
@@ -993,17 +994,20 @@ See the documentation on clip preferences for more details on how this is used w
 #define kOfxImageUnPreMultiplied "OfxImageAlphaUnPremultiplied"
 
 /** @brief Indicates whether an effect should premultiply or not on output.
-This provides clarification for generators or effect where input is kOfxImageOpaque.
+Added in version 1.5: This provides clarification for generators or effect where input is kOfxImageOpaque.
 
 - Type - string X 1
-- Property Set - a plugin  instance (read only)
-- Default - 0
+- Property Set - a plugin  instance (read only). Ideally this information should already be available at parameter creation stage.
```
garyo, line 1000: I don't think we should add comments like "Ideally". Just "should" or "must". Who would be helped by this change?

include/ofxImageEffect.h
```diff
@@ -993,17 +994,20 @@ See the documentation on clip preferences for more details on how this is used w
...
 Valid Values - This must be one of
 
 - kOfxImagePreMultiplied   - the image is premultiplied by its alpha
 - kOfxImageUnPreMultiplied - the image is unpremultiplied
 
-This Property defines an host level policy.
+This Property defines an host level policy. It is for practical purposes considered global in the scope of a Project.
```
garyo, line 1006: When you say "It is for practical purposes considered global" are you really saying that a host should (or must) not change its value? If so, do you mean it must be constant per plugin instance, constant for all instances of a given plugin, or what? And why? Why couldn't a host pass different values to different instances?

include/ofxImageEffect.h
```diff
@@ -993,17 +994,20 @@ See the documentation on clip preferences for more details on how this is used w
...
+
+Recommended Interpretation: If an host adds support for OfxImageEffectHostDesiredPremult, the host should set on the output the same value on output clip kOfxImageEffectPropPreMultiplication preference
```
garyo, line 1008: This needs some grammar work.

revisionfx: :) - upping the standard now

Also should not say Recommended but Expected here I guess

garyo:
This is now implemented, using `setOutputPremultiplication` and related.

Issue #46, opened 2017-10-12
Color-space handling


MrKepzie:
**Standard version:** 1.5 (Should be set as 1.5 milestone by @garyo)

**Subcommittee** 
Dennis Adams, Phil Barrett (Their Github user account (if any) should be set to assignees by @garyo)

## Standard Change Workflow
- [x] Create proposal as issue (you're doing this now!)
- [x] Tag this issue with `standard change` tag
- [x] Identify subcommittee: at least one plug-in vendor, and at least one host
- [x] Discuss the idea in this issue
- [ ] Write new or updated code and doc
- [ ] Publish updates as a pull request (ideally on a `feature/PROPOSAL-NAME` branch)
   - [ ] Make sure that PR references this issue number to keep them in sync
   - [ ] Discuss and review code in the PR
   - [ ] Meet all requirements below for accepting PR
- [ ] When subcommittee signs off and other members don't have any further review comments, 
      maintainer merges PR to master which closes PR and issue

## Requirements for accepting a standard change:
- [ ] Header files updated
- [ ] Documentation updated
- [ ] Release notes added
- [ ] Compatibility review completed
- [ ] Working code demonstrated with at least one host and one plugin
- [ ] At least two members sign off
- [ ] No further changes requested from membership

# Overview

There is little to no information about the colour space (tone curve and RGB chromaticities) in which image sample values are to be interpreted.

Historically there has been some assumption that the tone curve is probably linear, with a nominal black point at 0.0 and nominal white point at 1.0, but this is not specified and is often not the case.

Depending on the host, unmodified images presented to plugins may be linear, or "video" (typically Rec.1886, or a simple gamma curve), or "log" (e.g. Cineon log, or a modern log curve such as ARRI LogC), or "HDR" (e.g. Rec.2020) or any number of other curves. The primaries may be Rec.1886, or a larger gamut such as Rec.2020 or ARRI WideGamut. Plugins have no way of knowing what the numbers in the image represent. Similarly hosts have no way of knowing in which colour space(s) (if any) a plugin would like its inputs, nor the colour space of its output image.

Some plugins won't care what space the images are in. Some however need to know in order to make sense of the data. A lens flare, for example, needs to behave very differently (a) when the input data is linear with highlights well above 1.0 (b) when the input data is log-like with highlights preserved but much lower in the range, (c) when the input data is in a display-referred colour space where highlights have been rolled off to fit under 1.0.

[...]

PR #63, opened 2019-10-20
Creates ofxColourSpace.h

revisionfx:
Adds two levels of support for ColourSpace. See proposal at #46.

### DIFF
--- a/include/ofxColourSpace.h
+++ b/include/ofxColourSpace.h
@@ -4,7 +4,7 @@
 /*
 Software License :
 
-Copyright (c) 2003-2019, The Open Effects Association Ltd. All rights reserved.
+Copyright (c) 2003-2020, The Open Effects Association Ltd. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


include/ofxColourSpace.h
```diff
@@ -117,6 +117,15 @@ extern "C" {
 
 #define kOfxImageClipPropColourSpace "OfxImageClipPropColourSpace"
 
+/** @brief The approximate gamma value
+
+- Type - double X 1
+- Default: 1.0  - linear
+- Valid Values - can be any positive value, usual suspects are 2.2, 2.4, 2.6
+
+	This property is slightly orthogonal to the other ones in this property set. The main purpose is for a plug-in to internally linearize and delinearize non-destructively in a less arbitrary way.
```
garyo, line 126: This needs a lot more clarification. Is this the gamma that the plug-in should expect its input to be encoded with? Is it valid on input and output? During what calls is it valid?
In general I don't think there is enough info in this header to effectively implement this suite on either plugin or host side yet. Can some props be specified and others not? What's the expected calling/checking sequence on both sides? Maybe add a couple of worked examples or a table in the comments, so those will get into the doc.

revisionfx: Added some clarifications. Note this is exactly like we do in AE when user switches from project linearize versus color managed.  It does not imply any processing from host, it's just data for plugins.

PR #69, opened 2020-03-22
Update ofxsProcessing.H

RikshaDriver:
Bypass macro due to compiler errors thrown:

'(' : illegal token on right side of '::'

### DIFF
--- a/Support/Plugins/include/ofxsProcessing.H
+++ b/Support/Plugins/include/ofxsProcessing.H
@@ -134,10 +134,10 @@ namespace OFX {
             preProcess();
 
             // make sure there are at least 4096 pixels per CPU and at least 1 line par CPU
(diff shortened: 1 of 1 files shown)


devernay:
Hi Phil,
This is unfortunately necessary under windows if NOMINMAX is not defined, because Windows.h defines min() and max() as macros.

See also https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max

garyo:
Perhaps a better solution is to ensure `NOMINMAX` is defined on Windows. Without that it's hard to write modern code.

colin-alway:
I stumbled on this same issue trying to build in VS 2019. Alternative solution is to define the type:

`std::min<unsigned int>(...)`

garyo:
I think using `#define NOMINMAX` on Windows (either in your source or on the cmd line or in precompiled headers) is the correct solution to this, rather than convoluting the OFX header files. Closing. Please feel free to reopen if you have other ideas.

PR #85, opened 2021-08-31
Feature/param clip

revisionfx:


fxtech:
Is there documentation on how you would use this? What properties would you need to attach to the parameter? You would need at least a property that indicates the clip name for which this parameter binds. Or should the parameter name and clip name be the same?

fxtech:
I don't like the idea of requiring the parameter name to match the clip name.
Could we instead allow setting the kOfxClipName property on the parameter which specifies the clip/source name to bind to?

barretpj:
>
> I don't like the idea of requiring the parameter name to match the clip
> name.
> Could we instead allow setting the kOfxClipName property on the parameter
> which specifies the clip/source name to bind to?
>
That is what was suggested, as I read it. The wording isn't clear but I
think that's what it means:

"The "name" string value must match "name" string value of the optional
input clip"

revisionfx:
Yes this is so people can set optional / aux clips at the right place in their UI, and so it responds to show hide of a group... right now I have plugins with up to 7 inputs and putting all that at the top is rather confusing.

fxtech:
Some OFX hosts don't have the concept of a parameter "slot" for an input (they just use input noodles or what-not), so I assume it's ok to ignore this parameter type in that case.
Do you know of any OFX hosts that have this AE-style input assignment?

revisionfx:
Yes this is for non-nodal UI hosts so plug-in effects controls can look better, can be skipped if not supported. That just serves the same purpose as the input connector in DAG...  Since we don't have that right now people either stack them somehow on top of plugin controls or just support filter context single input effect. Note (different thing but should go together in UI)  in some hosts there is for single channel input an option menu that shows up for selecting channels.

Dithermaster:
Paul, I vaguely remember us suggesting this mechanism and there is some
chance it is implemented in Vegas Pro. You'd only see it for Effects that
have optional inputs, and we had a selector where you could select
the source as another track. I don't remember if it shipped or not. I did
just look and it is not implemented in Catalyst Edit.
///d@

fxtech:
Dennis- Yeah I do remember that, and I don't know either whether that 
was implemented in Vegas. But Vegas was the first host I thought of 
where this would make sense.

garyo:
This looks like a dup of #60

revisionfx:
As follow-up, this is useful for nodal UI as well:
- Enable/Disable or Show/Hide is useful for effects that have multiple inputs options - this could be mirrored to graph
For example one host allows only a minimum of inputs and they need to be connected for additional ones to be displayed

Issue #75, opened 2020-04-21
Overlay drawing API abstraction layer


fxtech:
# Open Effects Proposal for Standard Change

## Standard Change Workflow
- [x] Create proposal as issue (you're doing this now!)
- [x] Tag this issue with `standard change` tag
- [x] Identify subcommittee: at least one plug-in vendor, and at least one host
- [x] Discuss the idea in this issue
- [x] Write new or updated code and doc
- [x] Publish updates as a pull request (ideally on a `feature/PROPOSAL-NAME` branch)
   - [x] Make sure that PR references this issue number to keep them in sync
   - [x] Discuss and review code in the PR
   - [x] Meet all requirements below for accepting PR
- [x] When subcommittee signs off and other members don't have any further review comments, 
      maintainer merges PR to master which closes PR and issue

## Requirements for accepting a standard change:
- [x] Header files updated
- [x] Documentation updated
- [x] Release notes added
- [x] Compatibility review completed
- [x] Working code demonstrated with at least one host and one plugin
- [x] At least two members sign off
- [x] No further changes requested from membership

## Summary

A drawing utility suite to abstract away required use of OpenGL.

## Motivation

With OpenGL "going away" and new GPU and UI layers being used, we should provide an abstraction layer so plugins can draw on-screen controls or custom parameter UIs without requiring OpenGL.

## Problem

Demarcation, that's the problem!

## Impact

This suite is initially optional for both host and plugin use, and mostly embodied in a new header. If a host provides the suite, the plugin should make use of it, otherwise it is free to fall back to OpenGL. However, it should be considered as a required feature in future OFX versions, since OpenGL is slowly being deprecated.

One new POD type should be added to ofxCore.h.

## Documentation Impact

New documentation will have to be integrated.

## Stakeholders

Plugin writers will benefit as they will no longer be required to use OpenGL for their interacts.
Hosts will benefit as they will not be required to support OpenGL at all.

barretpj:
Using this instead of letting plugins use GL directly would be a big win for Baselight - currently we have to intercept all interact GL calls to route them to our separate rendering process which is often on a separate system. It works but I'd be delighted not to have to do it!

peterl-bmd:
Hosts will require some means to determine if a plugin is going to exclusively use the draw suite.  If we are not running in OpenGL, we cannot call interacts for a plugin that is going to make GL calls, without likely crashes/undefined behaviour.

[...]

PR #86, opened 2021-08-31
Create ofxDrawSuite.h

revisionfx:
Adding Paul Miller's Overlay Draw Suite proposal for review

### DIFF
--- a/include/ofxDrawSuite.h
+++ b/include/ofxDrawSuite.h
@@ -94,15 +94,20 @@ typedef enum OfxDrawLineStipplePattern
 } OfxDrawLineStipplePattern;
 	
 /** @brief Defines valid values for OfxDrawSuiteV1::draw */
+
 typedef enum OfxDrawPrimitive
 {
 	kOfxDrawPrimitiveLines,
 	kOfxDrawPrimitiveLineStrip,
 	kOfxDrawPrimitiveLineLoop,
-	kOfxDrawPrimitiveRectangles,
+	kOfxDrawPrimitiveRectangle, /* 2 points*/
 	kOfxDrawPrimitivePolygon,
 	kOfxDrawPrimitiveEllipse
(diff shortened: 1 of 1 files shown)


revisionfx:
Issues 74 and 75 ?

revisionfx (commented):
Should Polygon and Ellipse be plural like Lines and Rectangles?
(or the opposite - something consistent)

fxtech:
> Should Polygon and Ellipse be plural like Lines and Rectangles?
> (or the opposite - something consistent)

Probably not for Polygon - otherwise how would you know how many verts each one has?
I tried to make it analogous to glBegin/glEnd.

revisionfx:
Paul, ok makes sense for Polygon
Maybe Rectangles should be singular? :)

paul-miller-borisfx:
> Paul, ok makes sense for Polygon
> Maybe Rectangles should be singular? :)

Rectangles was meant as an analogue for GL_QUADS. And there isn't a GL_QUAD.

revisionfx:
I see, then maybe it should be called Quads :)
Already two developers got it wrong  - treated this as glRectf   (2 coords)
And if quads then specify if CW or CCW in comment?

fxtech:
> I see, then maybe it should be called Quads :)
> Already two developers got it wrong - treated this as glRectf (2 coords)
> And if quads then specify if CW or CCW in comment?

How about we replace it with RECTANGLE, and just require 2 coordines? That'll remove any ambiguity.

revisionfx:
OK for 2 points rect for me

revisionfx:
Paul, I updated Rectangles for Rectangle - and added an #if 0 for now if someone wants to support existing version in the wild using the original version.

garyo:
Merged onto branch for review; please continue discussion in `feature/overlayDraw` branch or in issue #75

PR #92, opened 2022-04-25
Add CMake & Conan support

Iswenzz:
Hello, This PR allows building the OpenFX library with CMake and Conan, and simplifies the creation of plugins (Using previous PR https://github.com/ofxa/openfx/pull/40).


revisionfx:
add it in support folder and pull request it

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+# Flags
+if(MSVC)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWINDOWS -DNOMINMAX")
+else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wno-deprecated -Wno-deprecated-declarations")
+endif()
+
+# Conan
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
```
garyo, line 17: This seems to require conan; plain cmake fails here with "include could not find requested file" conanbuildinfo.cmake. Is that your intent here? I think it would be better to allow a plain cmake build if possible.

garyo (commented):
I was able to build this on Mac as follows:
```sh
mkdir build; cd build
conan install .. --build=expat
cmake install ..
make
```
That produced `build/lib/libOfxHost.a` and `build/lib/libOfxSupport.a`. I have not tried those in an actual OFX plugin or host build.

garyo:
This will be merged with additions in #109. Thanks!

PR #93, opened 2022-07-20
Replace randomGenerator.cpp/H with C++11 <random>

garyo:
This improves the random number quality, as well as removing problematically
licensed (LGPL) code from the example.


revisionfx (approved):
See my comment  (question) - does not affect being an example

PR #97, opened 2022-09-01
Enable host to know if a plug-in uses OfxDrawSuite to draw overlays

MrKepzie:
This addresses #75  where the host cannot know if a plug-in will use the new OfxDrawSuite to implement the draw action of the Overlay Interact. 
Setting a new kOfxImageEffectPluginPropOverlayInteractV2 is a good way to indicate the host that the plug-in's overlay interact uses the draw suite.

### DIFF
--- a/include/ofxImageEffect.h
+++ b/include/ofxImageEffect.h
@@ -827,6 +827,19 @@ The entry point pointed to must be one that handles custom interaction actions.
 */
 #define kOfxImageEffectPluginPropOverlayInteractV1 "OfxImageEffectPluginPropOverlayInteractV1"
 
+/** @brief Sets the entry for an effect's overlay interaction. Unlike
+ kOfxImageEffectPluginPropOverlayInteractV1, the overlay interact in the plug-in is expected
+ to implement the kOfxInteractActionDraw using the OfxDrawSuiteV1.
+ 
+   - Type - pointer X 1
+   - Property Set - plugin descriptor (read/write)
+   - Default - NULL
+   - Valid Values - must point to an ::OfxPluginEntryPoint
+
+The entry point pointed to must be one that handles custom interaction actions.
+*/
+#define kOfxImageEffectPluginPropOverlayInteractV2 "OfxImageEffectPluginPropOverlayInteractV2"
+
 /** @brief Indicates whether a plugin or host support multiple resolution images.
 
    - Type - int X 1


barretpj:
OK so we're saying that `OfxImageEffectPluginPropOverlayInteractV1` is the entry point for an OpenGL-drawing interact, and `OfxImageEffectPluginPropOverlayInteractV2` is the entry point for a DrawSuite-drawing interact.

In that case, the header docs for `OfxImageEffectPluginPropOverlayInteractV1` need to make it explicit that it cannot use DrawSuite.

revisionfx:
Phil, this would break current implementations (4 or 5 I know of) that use new suite with InteractV1 and can do book-keeping if plugin fetches draw suite... Perhaps if V2 is supported expect host to always use drawSuite. If V2 is not supported, do as currently if drawsuite is not supported expect OpenGL calls... else expect drawsuite and test what happens in app before shipping...

barretpj:
I can't see how we can accurately implement this change (giving the host a way to know how the plugin will draw overlays) without breaking existing plugins that clearly won't have implemented any new mechanism.

To allow existing plugins to remain unchanged, we'd be saying a plugin must be able to use either draw method in the V1 call, with no additional properties, and hosts wouldn't be able to tell. And if the host only offers DrawSuite it wouldn't be able to know that the plugin couldn't draw its interact.

revisionfx:
* for plug-ins as they are used to Overlay not supported or supported, it's simple to check if new suite is supported
* for a host we could if it helps assume that draw suite is hand-shaked first before Interact otherwise host does not know what pointer to give you. -> If so, this should maybe be added in header doc of DrawSuite

If plugin can support V2 and V1, they ask for V2 and if not supported, then ask for V1 and maintain that pointer
We have a number of hosts supporting new suite via V1 and legacy plugin so best we can do I think is:
V2 property would be DrawSuite only ? A host can decide not to support V1 (as V1 might work with OpenGL and they might draw all their UI in DIrectX or Metal or whatever kids do these days)

fxtech:
Implemented this check in Silhouette 2022.5.

fxtech (changes requested):
Add a note about the requirement of OpenGL for using kOfxImageEffectPluginPropOverlayInteractV1

barretpj:
Agreed, if the plugin has a V2 entry point then that's DrawSuite only. V1 can be DrawSuite if the suite is available from the host, but it must support GL.
Host should use a plugin's V2 if it's present.
We also need to do this for `kOfxParamPropInteractV1`/`V2` I assume.

revisionfx:
> 

Why do we need this for Interact suite too?

barretpj:
> Why do we need this for Interact suite too?

The primary need is in the interact suite. My comment above is that we need to do it for parameters with custom-drawn UI as well, because a non-GL host would presumably want them to draw with the DrawSuite too. (No idea whether any hosts support custom-UI params these days though?)

revisionfx:
> 

Good point, added to wiki as open question
https://wiki.aswf.io/display/OFX/Roadmap

fxtech (approved):
This is all good.
I think it would be worthwhile to add some clarifying text however that firms up that V1 interacts are meant to use OpenGL.

barretpj:
As discussed at yesterday's TSC I suggest the following comment changes

```diff --git a/openfx/include/ofxParam.h b/openfx/include/ofxParam.h
index 7483d9242bf..4f12ada94d7 100644
--- a/openfx/include/ofxParam.h
+++ b/openfx/include/ofxParam.h
@@ -144,6 +144,9 @@ These are the list of properties used by the parameters suite.
     - Type - int X 1
     - Property Set - host descriptor (read only)
     - Valid Values - 0 or 1
+
+Currently custom interacts for parameters can only be drawn using OpenGL.
+APIs will be added later to support using the new Draw Suite.
 */
 #define kOfxParamHostPropSupportsCustomInteract "OfxParamHostPropSupportsCustomInteract"
 
@@ -200,6 +203,9 @@ See \ref ParametersInterfacesPagedLayouts for more details.
     - Valid Values -  must point to a OfxPluginEntryPoint
 
 If set, the parameter's normal interface is replaced completely by the interact gui.
+
+Currently custom interacts for parameters can only be drawn using OpenGL.
+APIs will be added later to support using the new Draw Suite.
 */
 #define kOfxParamPropInteractV1 "OfxParamPropInteractV1"
 
```

revisionfx (approved):

garyo (approved):
Looks good to me; I will create a followup PR with the doc changes discussed in this PR.

PR #105, opened 2023-01-04
Doc: clarify that custom param UIs must use OpenGL, not DrawSuite

garyo:


garyo:
Tried without DCO -- will repost.

PR #107, opened 2023-01-04
Move OpenGL suite & CUDA/Metal/OpenCL into new ofxGPURender.h

garyo:
The new props should not have been in the main `ofxImageEffect.h`; moved them into the new file, along with the content of `ofxOpenGLRender.h` which now just includes the new `ofxGPURender.h`.

I also added `#pragma once` to some of the headers and fixed a missing `#include`. These should be done globally, IMHO.
Also improved the doc for some of the new props.


revisionfx (approved):
I have other requests to add after this, like a way to switch dynamically if we want gpu images or not... will add in a second round, not to make this contingent on other things. First step was having a generic GPU render header file.

revisionfx (approved):
I check with BMD, apparently the latest release includes our changes (their SDK dropped with app)...

garyo:
Is there any reason to wait on merging this?

Dithermaster:
It doesn't have the OpenCL image parts (combined header) just the BMD buffer part. If this can be added later, then OK to merge. If you want it included, then I need to add it.

garyo:
Thanks for catching the missing OpenCL! It's probably simplest for me to merge this and then you can fix up OpenCL (rather than trying to patch my PR).

PR #109, opened 2023-01-05
Cmake support

garyo:
This merges #92 and adds building plugins as well as the support libs.

- Add `scripts/build-cmake.sh` to build everything with conan and cmake (which must be installed)
- Build all example plugins (both with and without the support lib)
- Set up correct directory structure for plugins
- Add "install" cmake target to install plugins locally for testing
- Allow verbose building with "-v"
- Fix MacOSX build to build as bundle
- Add MacOSX universal architectures (x86_64 & arm64)
- Switch conan to cmake_find_package for more targeted lib linking; this way only the plugins that need OpenGL link with it.

Tested on MacOS; not yet tested on Windows and Linux.


garyo:
Note that this should also enable us to push the conan recipe to the conancenter repository, so OFX host and plugin vendors will be able to pull OFX directly from conancenter.

john-paulsmith:
I just tested this and the instructions didn't work because Conan 2.0.0 has just been released, which contains breaking changes. I rolled back to Conan 1.5.9 and got further but wasn't able to build the plug-ins. Just to be awkward I was using Linux on aarch64 which might have exposed some issues. There was no expat package available so I had to use the --build=expat option to conan install, then found HostSupport/include/ofxhBinary.h couldn't handle aarch64. I'll work through the other issues to see if I can get it to build.

Main question though, do we care about the fact that this doesn't work with Conan v2?

garyo:
Thanks for doing the test! Good question re: conan 2.0. I suggest we push the conan 1.x version now and add a ticket for conan v2 support, since everyone is currently still using conan v1.
I will add `--build=missing` to the instructions, which will handle any missing dependencies.
If `ofxhBinary.h` has architecture-specific issues, those are perhaps worth a separate PR.

cmake/OpenFX.cmake
```diff
@@ -0,0 +1,53 @@
+if(APPLE)
+  set(PLUGINDIR "/Library/OFX/Plugins")
+  set(ARCHDIR "MacOS-x86-64")
```
john-paulsmith, line 3: Change this to `MacOS`, the universal binary location. Plug-ins built with this script on an Apple Silicon Mac come out as universal so don't work unless the binaries are in MacOS.

readme.md
```diff
@@ -39,6 +40,40 @@ The Open Effects Association (OFX), a non-profit organization, develops and prom
 
 The Association's initial focus is to improve the OpenFX image processing plug-in standard. This goal of this standard is to reduce development effort and support needed for plug-ins across different compositing and editing host platforms.
 
+# Building
```
john-paulsmith, line 43: You could refer to `scripts/build-cmake.sh` instead of including the details here.

john-paulsmith (changes requested):
I have highlighted a few doc issues in the readme and functional issues in `build-cmake.sh`. With my suggested changes to the script, the plug-ins build and operate correctly on macOS.

Iswenzz:
> Note that this should also enable us to push the conan recipe to the conancenter repository, so OFX host and plugin vendors will be able to pull OFX directly from conancenter.

I made a conan recipe a few months ago with my first PR, if you wanna check it out:
https://github.com/conan-io/conan-center-index/pull/10512

garyo:
Thanks for doing that recipe, @Iswenzz ! Would you mind seeing how it could be merged with the recipe in this PR? Any suggestions would be helpful.

And of course since conan 2.0 was just released, if you could help us port the recipe to 2.0 that would be awesome!

acolwell:
@garyo I just uploaded #113 which gets what you started here working with Conan 1.x and Conan 2.x. I haven't tested it on a Mac since I don't have access to one. Let me know if you need any changes.

garyo:
This should now be complete, with Conan 1.59 and 2.x support thanks to @acolwell in #113 which is now merged into this branch. @john-paulsmith please take a look.

scripts/build-cmake.sh
```diff
@@ -4,36 +4,54 @@ set -e
...
+cmake --preset ${PRESET_NAME} -DBUILD_EXAMPLE_PLUGINS=TRUE
 
 # Do the build
-cmake --build $BUILDDIR $VERBOSE
+cmake --build ${CMAKE_BUILD_DIR} --config $BUILDTYPE $VERBOSE
 
-# install (or you could just do "cmake --build $BUILDDIR --target install")
-cmake --install $BUILDDIR
+# install (or you could just do "cmake --build $BUILDDIR --config $BUILDTYPE --target install")
+#cmake --install build/ --config $BUILDTYPE
```
john-paulsmith: With this line commented out, obviously the plug-ins don't get installed anymore. It should be changed to `cmake --install ${CMAKE_BUILD_DIR} --config $BUILDTYPE`. The comment above about the alternative command should also be changed to incorporate ${CMAKE_BUILD_DIR} instead of $BUILDDIR due to other changes in the script.

john-paulsmith: Apart from this, everything looks good, @garyo if you can make this change I will complete the review.

garyo: I think @acolwell commented this out because it'll fail the build unless the OFX install dir exists and is world-writable. When I create the CI build script, I can make sure that exists before running this script, but I think for users we should just tell them to run the `cmake --install` command as root (or admin, for Windows). Or do you think we should just fail the build there and add an error message telling them to create the install dir and ensure it's writable by them? Or perhaps check for this up front?

I agree about fixing the build dir name.

garyo:
I've also discovered that with Conan 1.59.0 on Windows, doing two builds in a row with `scripts/build-cmake.sh` fails. The first one succeeds but the second fails in the `conan --install` step with this error:
```
Building OpenFX Release in build/ with conan platform default generator
+ conan install -s build_type=Release --build=missing .
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=17
os=Windows
os_build=Windows
[options]
[build_requires]
[env]

Preset CMake variables:

  CMAKE_POLICY_DEFAULT_CMP0091="NEW"
  CMAKE_TOOLCHAIN_FILE:FILEPATH="c:\Users\garyo\src\OpenFX\build\generators\conan_toolchain.cmake"

CMake Error: Error: generator platform: 
Does not match the platform used previously: x64
Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
```

scripts/build-cmake.sh
john-paulsmith: For a developer new to the project, having the install phase run automatically is very helpful. I would rather see the build fail with a clear permissions error than skip it and leave that new developer scratching their head about how to get the plug-ins to show up in a host. Checking up front would get my vote.

acolwell: Yes. I disabled it because it failed and I had concerns about it writing to a global directory like this. I also disagree with combining the build and install steps here.  "make" then "make install" is a common pattern in projects and allows the developer to explicitly decide whether to install the package in a default location or not.

As a new developer to this project, I was quite surprised that this script was trying to install into a global location. Obviously this is your project and you can do what you will. I just figured I'd respond with my reasoning in the hopes that it would be helpful.

acolwell:
> I've also discovered that with Conan 1.59.0 on Windows, doing two builds in a row with `scripts/build-cmake.sh` fails. The first one succeeds but the second fails in the `conan --install` step with this error:
> 
I believe this is because "conan install" is really only meant to be called once when setting up a build. The idea is to download/build all the deps and generate anything needed for a build environment. You could update the code to conditionally run it based on the presence of the conan_toolchain.cmake file or CMakeCache.txt, but then it could potentially not run when the user changes deps in the conanfile.py.  I suspect you'll need to figure out the best tradeoff between "run a single command and everything 'just works' for the most common cases" vs "Run these commands to setup the build, and then these other ones to build and install." 

I was under the impression this script was intended just as an example and not intended and a developer's primary workflow.

scripts/build-cmake.sh
john-paulsmith: I agree "make" then "make install" is well understood. Maybe the thing to do is remove the install step from build-cmake.sh (or enable it via a command-line arg) then document the cmake --install command in readme.md.

garyo:
I think it is (and should be) a good example, and people may build workflows from it. It's also part of the "out of box experience" so I'd like that to be smooth, to give new users confidence.

Interestingly, I can re-run `conan install -s build_type=Release --build=missing .` from my shell with no errors. It's only a problem if I re-run it via the `build-cmake.sh` script, which likely means it's just something set in the environment of that script. I'll poke at it.

I will separate out the install part, as discussed above.

john-paulsmith (approved):

PR #110, opened 2023-01-05
Update all copyrights to new ASWF spec

garyo:


Documentation/sources/Guide/Code/Example1/basics.cpp
```diff
@@ -1,31 +1,6 @@
...
-ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
-(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
-ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-*/
+// Copyright Contributors to the OpenFX project.
+// SPDX-License-Identifier: BSD-3-Clause
+// https://github.com/AcademySoftwareFoundation/OpenFx
```
jmertic, line 3: This line isn't needed

jmertic:
Two bits:

- You don't need the 'https://github.com/AcademySoftwareFoundation/OpenFx' line in the header
- Please update the LICENSE file to the standard BSD-3 clause text

Otherwise looks good!

Documentation/sources/Guide/Code/Example1/basics.cpp
revisionfx: Maybe be it's Copyright that is redundant, so it works everywhere they have copyright laws
reference:  https://liferay.dev/blogs/-/blogs/how-and-why-to-properly-write-copyright-statements-in-your-code
(c) Contributors to the OpenFX project

garyo: I will remove the URL everywhere. Agree that's not needed (I got it from another ASWF project). As for © vs Copyright, let's go with the ASWF standard unless there's a good reason to be different. The ASWF lawyers will be the ones defending a copyright suit if that ever happened.

garyo:
> Please update the LICENSE file to the standard BSD-3 clause text

Top-level LICENSE.md looks like the correct BSD-3-Clause license to me (https://opensource.org/licenses/BSD-3-Clause) -- did I miss something?

I do see that `Support/LICENSE` file was a bit wrong, so I updated that to be correct BSD-3-Clause and deleted some old unneeded license header text (duplicating the license terms already specified by the license clause).

john-paulsmith:
Following on from @revisionfx's comment on the issue, our project has a mixed copyright, so I think the copyright message should be `Copyright OpenFX and contributors to the OpenFX project`. To avoid confusion, we'd be better off using this in all files rather than just old files.

garyo:
I will implement whatever the consensus is here, but I don't have enough knowledge to make a decision. 
I'm not sure who the copyright holder is in "Copyright OpenFX" -- the OpenFX LLC series?

john-paulsmith:
> I'm not sure who the copyright holder is in "Copyright OpenFX" -- the OpenFX LLC series?

Yes, according to the legal docs it's "OpenFX a Series of LF Projects, LLC", and I'm assuming it's reasonable to abbreviate that to OpenFX. Maybe that's not a reasonable assumption, in which case we could include the full name. @jmertic do you have any thoughts on that?

garyo:
Any further thoughts?

garyo:
According to the [IP transfer agreement](https://www.dropbox.com/home/OFXA/Contracts?preview=IP-transfer-to-ASWF-2022.pdf), the official Assignee is actually "LF Projects, LLC".

john-paulsmith:
> According to the [IP transfer agreement](https://www.dropbox.com/home/OFXA/Contracts?preview=IP-transfer-to-ASWF-2022.pdf), the official Assignee is actually "LF Projects, LLC".

Clause 2.1 of the agreement splits the assignments. Only trademarks were assigned to LF Projects, the copyright and all other IPR was assigned to the Series.

jmertic:
> > I'm not sure who the copyright holder is in "Copyright OpenFX" -- the OpenFX LLC series?
> 
> Yes, according to the legal docs it's "OpenFX a Series of LF Projects, LLC", and I'm assuming it's reasonable to abbreviate that to OpenFX. Maybe that's not a reasonable assumption, in which case we could include the full name. @jmertic do you have any thoughts on that?

Yes - abbreviating is fine.

john-paulsmith (approved):

PR #112, opened 2023-02-17
Remove the website from main

john-paulsmith:
The website and related files are not relevant for most developers so it would be better to not include them when checking out main. I think this is doubly true because the website is a snapshot which we don't expect to edit again. This PR would remove it from main and leave it on the website branch.


garyo (approved):
Makes sense. LGTM.

PR #115, opened 2023-03-18
Replace deprecated auto_ptr with unique_ptr.

acolwell:
Replaced all std::auto_ptr usage with std::unique_ptr usage. auto_ptr was deprecated in C++11 and was removed in C++17. I tried to make minimal changes that would preserve behavior and avoid changing too many function signatures.


garyo (approved):
Looks good to me; builds on my Mac (as long as I update the Makefiles to c++11; see other CMake/conan PRs for that)

PR #119, opened 2023-05-08
Add github action to build all in CI

garyo:


garyo:
@Iswenzz I'm trying to get a CI build working. I have it working fine on ubuntu, windows and mac, so now I'm working on the ASWF VFX Reference Platform -- see the last commit in this repo. I'm stuck on how to invoke cmake after running `conan install` because that platform (CentOS 7) uses an older cmake, v3.19.3 and that gives some errors. I think we should ensure that OpenFX builds on that platform; any thoughts on how to make that work? 
I filed a question with conan explaining the details: https://github.com/conan-io/conan/issues/13877 (so I won't repeat them here) and I thought maybe you have a quick solution.

garyo:
Actually I think I figured it out; the top-level `CMakeLists.txt` was hard setting `CMAKE_MODULE_PATH` which was overriding the path set up in the conan toolchain. Using list append there seems to work.

PR #122, opened 2023-06-15
Fix example makefiles to use proper path to symbol files

garyo:
The CMake builds are working now, but the examples were still using an old path to the symbol files.


john-paulsmith (changes requested):
The changes now point to the correct path but the argument syntax is incorrect on macOS. Linux is fine.

john-paulsmith:
In my testing, I found that the symbol files are unnecessary on macOS and Linux - probably on Windows too but I'm not set up to test that easily. We set `-fvisibility=hidden` then in the code of our examples we include `__attribute__((visibility("default")))` on `OfxGetPlugin` and `OFXGetNumberOfPlugins`, which overrides the command-line arg and exports these symbols.

Not specifically related to this commit, but the CMake project has the same error in the macOS arguments. I will open a new PR to set visibility=hidden in a cross-platform way there.

garyo:
I have this working, but I see you've also got something similar in #141, @john-paulsmith . I'll merge your changes there into this or vice versa because this also has some only tangentially-related build fixes.

Issue #124, opened 2023-06-17
Add VFX Platform 2023 to CI build matrix


jfpanisset:
There is now a VFX Platform 2023 version of the ci-base aswfdocker build container available:

* [aswf/ci-base:2023](https://hub.docker.com/layers/aswf/ci-base/2023/images/sha256-4d96aea886237b266ead9b18b164c5b5ebb797374f22829d9872ed7e45744f0c?context=explore)

It could be added to the CI build matrix:

https://github.com/AcademySoftwareFoundation/openfx/blob/95a1179ac20357f165d89906ad1f49acc89fcede/.github/workflows/build.yml#L37

Also since OpenFX is now part of ASWF, are there required build pre-requisites which would be needed and would justify a specific `ci-openfx` build container, or could just be rolled into `ci-base`, or do you currently have everything you need?

PR #125, opened 2023-06-19
CI: Add build on VFX Reference Platform 2023

garyo:
See #124

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
           - name_prefix: Linux CentOS 7 VFX CY2021
             release_prefix: linux-vfx2021
             ostype: linux
-            distro: centos7
+            aswfdockerbuild: true
             os: ubuntu-latest
             container: aswf/ci-base:2021
             vfx-cy: 2021
@@ -32,7 +32,7 @@ jobs:
           - name_prefix: Linux CentOS 7 VFX CY2022
             release_prefix: linux-vfx2022
             ostype: linux
-            distro: centos7
+            aswfdockerbuild: true
             os: ubuntu-latest
             container: aswf/ci-base:2022
             vfx-cy: 2022
@@ -43,10 +43,10 @@ jobs:
             cxx-compiler: clang++
             cc-compiler: clang
             compiler-desc: Clang
-          - name_prefix: Linux CentOS 7 VFX CY2023
+          - name_prefix: Linux Rocky 8 VFX CY2023
             release_prefix: linux-vfx2023
             ostype: linux
-            distro: centos7
+            aswfdockerbuild: true
             os: ubuntu-latest
             container: aswf/ci-base:2023
             vfx-cy: 2023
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -43,6 +43,20 @@ jobs:
             cxx-compiler: clang++
             cc-compiler: clang
             compiler-desc: Clang
+          - name_prefix: Linux CentOS 7 VFX CY2023
+            release_prefix: linux-vfx2023
+            ostype: linux
+            distro: centos7
```
jfpanisset, line 49: The 2023 aswf-docker containers are based on Rocky Linux 8 (8.8 as of the time they were published). Looks like the only reference to the `distro` symbol is to check whether it is ubuntu or not, which really seems to mean whether the build is in a `aswf-docker` container or not. Changing this value to `rocky8` would work, but perhaps the symbol should be called `aswfdockerbuild` and be a boolean?

.github/workflows/build.yml
```diff
@@ -43,6 +43,20 @@ jobs:
             cxx-compiler: clang++
             cc-compiler: clang
             compiler-desc: Clang
+          - name_prefix: Linux CentOS 7 VFX CY2023
```
jfpanisset, line 46: Name should probably be `Linux Rocky 8 VFX CY2023`

.github/workflows/build.yml
```diff
@@ -43,6 +43,20 @@ jobs:
...
+            release_prefix: linux-vfx2023
+            ostype: linux
+            distro: centos7
+            os: ubuntu-latest
```
jfpanisset, line 50: This is also present for the 2021 and 2022 aswf-docker containers, and seems to get written to a `OSNAME` env var.

.github/workflows/build.yml
```diff
@@ -43,6 +43,20 @@ jobs:
...
+            container: aswf/ci-base:2023
+            vfx-cy: 2023
+            has_cmake_presets: false
+            buildtype: Release
+            conan_version: 2.0.4
```
jfpanisset, line 55: `aswf-docker` containers already have a version of Conan, currently 1.58.0, but it probably makes sense to use your own version across all builds, especially given the flux created by the transition from 1.x to 2.x

jfpanisset (approved):
Should be good to go in as is to test builds, and address comments later if relevant.

.github/workflows/build.yml
garyo: Fixed.

.github/workflows/build.yml
garyo: Not sure what fix this comment is suggesting. I'm not sure OSNAME is used, but not worth removing at the moment. The `os` itself is used to select the OS to run the CI build on.

jfpanisset: I think this comment is obsolete, your second commit got rid of the confusing `distro: centos7` label which was confusing me with `os: ubuntu-latest` but of course the `os:` label is for the runner itself, and the `distro:` was meant for what's inside the container running the build.

PR #140, opened 2024-02-02
Add OpenCL Image support to the updated GPU rendering extension:

SonyDennisAdams:
- Modify ofxGPURender.h to support OpenCL Images
- Add kOfxImageEffectActionBeginSequenceRender and kOfxImageEffectActionEndSequenceRender as required actions for some CUDA/Metal/OpenCL properties where they were not listed
- Also make documentation edits in the same header
- Modify GPUGain example to also support OpenCL Images and show how to disambiguate in the case where a plug-in supports both OpenCL Buffers and OpenCL Images

### DIFF
--- a/Examples/GPUGain/GPUGain.cpp
+++ b/Examples/GPUGain/GPUGain.cpp
@@ -12,7 +12,7 @@
 
 #define kPluginName "GPU Gain"
 #define kPluginGrouping "OFX Example"
-#define kPluginDescription "Apply separate RGB gain adjustments to each channels; CUDA/OpenCL/Metal"
+#define kPluginDescription "Apply separate RGB gain adjustments to each channels; CUDA/OpenCL Buffers/OpenCL Images/Metal"
 #define kPluginIdentifier "com.OpenFXSample.GPUGain"
 #define kPluginVersionMajor 1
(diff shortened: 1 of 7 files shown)


fxtech:
When I tried to build MacOS on the gpu-gain-example branch I get this error:
```
CMake Error at CMakeLists.txt:51 (find_package):
  By not providing "Findopengl_system.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "opengl_system", but CMake did not find one.

  Could not find a package configuration file provided by "opengl_system"
  with any of the following names:

    opengl_systemConfig.cmake
    opengl_system-config.cmake

  Add the installation prefix of "opengl_system" to CMAKE_PREFIX_PATH or set
  "opengl_system_DIR" to a directory containing one of the above files.  If
  "opengl_system" provides a separate development package or SDK, be sure it
  has been installed.
```

revisionfx:
Would it be possible to add a property that states if the OpenCL Image is backed by an OpenGL texture? 
There is one host for example that supports OpenCL but not OpenGL...
Also can a plugin report based on parameter state it does not support OpenCL now just OpenGL... ?

garyo (approved):
LGTM!

PR #141, opened 2024-02-07
Remove symbol files as these are no longer needed

john-paulsmith:
Both the CMake and Makefile projects included compiler arguments to set symbol visibility to hidden and expose only the `OfxGetNumberOfPlugins` and `OfxGetPlugin` symbols. The CMake project has been changed to use built-in functionality to control symbol visibility, reducing the amount of platform-specific code that needs to be maintained.

Furthermore, all our examples use `__attribute__((visibility("default")))` to expose necessary symbols, rendering the symbol files unnecessary, so all references to the symbol files, and the files themselves have been removed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,6 @@ find_package(EXPAT)
 find_package(opengl_system)
 
 # Macros
-set(OFX_SUPPORT_SYMBOLS_DIR ${PROJECT_SOURCE_DIR}/symbols)
 include(OpenFX)
 set_property(GLOBAL PROPERTY OFX_PROJECT_SOURCE_DIR "${PROJECT_SOURCE_DIR}")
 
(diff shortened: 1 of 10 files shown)


john-paulsmith:
I have only tested this complete changeset on macOS, although I verified the concept on Linux too. Comments from a Windows perspective would be appreciated, I believe Windows symbol visibility is hidden by default and the use of __declspec(dllexport) will be exposing the correct symbols already.

Is there any reason to keep the symbol files around for plug-in/host build scripts?

john-paulsmith:
If this PR is merged, I believe #122 can be discarded.

garyo (approved):
LGTM -- I've pulled these changes into the work I did in #122 since there were some related changes in there to make sure the CI builds work.

garyo:
Closing; all in #122 now.

PR #150, opened 2024-03-29
Fix missing GPU include

jgoueslard:
Some variables like kOfxImageEffectPropOpenCLRenderSupported are used in ofxsImageEffect and are declared in ofxGPURender.h. So this file should always be included even if we build without OFX_SUPPORTS_OPENGLRENDER.

### DIFF
--- a/Support/Library/ofxsImageEffect.cpp
+++ b/Support/Library/ofxsImageEffect.cpp
@@ -10,9 +10,7 @@
 #include <iostream>
 #endif
 #include <stdexcept>
-#ifdef OFX_SUPPORTS_OPENGLRENDER
 #include "ofxGPURender.h"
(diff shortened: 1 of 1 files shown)


garyo (approved):
LGTM!

garyo:
The Windows CI failure is nothing to do with this change, it's something wrong with the CUDA installer github action. Just FYI.

garyo:
Note: we now require contributor sign-off on all commits. If you could re-push this branch with an amended commit message including "Signed-off-by: Author Name <authoremail@example.com>" then we can merge it. You can also do this automatically by using the -s flag (i.e., `git commit -s`).

jgoueslard:
Hi,
It's done, you can merge :)

PR #155, opened 2024-04-01
Add section about private extension reverse domain naming

SonyDennisAdams:
Add section about private extension reverse domain naming


garyo (approved):
LGTM again! :-)

PR #162, opened 2024-05-09
WIP: Colorspace example

garyo:


revisionfx:
@john-paulsmith
No one has answered me about having a role_only mode (e.g. "scene_linear", "sdr" )  so reasking here.
@garyo says it was facultative, so can we just have a special mode for that only (no OCIO strings to set). 
All I want in generic case is to query in render call is which of:   linear, sdr, hdr, log, data  - nothing else and I assume if I don't set in clip preferences log, then I won't get that.

garyo:
This has now been merged into PR #103. Closing.

PR #166, opened 2024-08-28
Add property checker/metadata system and C++ helpers

garyo:
* `include/ofx-props.yml` is the property reference/metadata
* `scripts/gen-props.py` is the checker and source generator

All props are included in the YML definition, and the script generates C++ files with all the property metadata and suite assignments. The C++ files aren't yet checked in, but you can build them easily: `python scripts/gen-props.py -v` and then look in `include/gen_*.hxx`.


include/ofxPropsMetadata.h
```diff
@@ -0,0 +1,224 @@
...
+struct PropsMetadata {
+  std::string name;
+  std::vector<PropType> types;
+  int dimension;
+  Writable writable;
+  bool host_optional;
+  std::vector<const char *> values; // for enums
+};
+
+const std::vector<struct PropsMetadata> props_metadata {
```
abizeauMaxon, line 43: Since this it auto generated, you could consider `std::array<struct PropsMetadata, N=FeedByScrip>`.

If this header is intended to be use more C++, you should consider having this `std::array` as constexpr.

maybe:

```
#if defined(__cplusplus)
#define CONST_SPECIFIER constexpr
#else
#define CONST_SPECIFIER const
#endif
```

include/ofxPropsMetadata.h
```diff
@@ -0,0 +1,224 @@
...
+   All
+};
+
+struct PropsMetadata {
+  std::string name;
+  std::vector<PropType> types;
+  int dimension;
+  Writable writable;
+  bool host_optional;
+  std::vector<const char *> values; // for enums
```
abizeauMaxon, line 40: You should be consistant, if you use `std::string` for name, you should also use `std::vector<std::string>`

include/ofxPropsBySet.h
```diff
@@ -0,0 +1,733 @@
...
+#include "ofxGPURender.h"
+#include "ofxColour.h"
+#include "ofxDrawSuite.h"
+#include "ofxParametricParam.h"
+#include "ofxKeySyms.h"
+// #include "ofxOld.h"
+
+namespace OpenFX {
+// Properties for property sets
+const std::map<const char *, std::vector<const char *>> prop_sets {
```
abizeauMaxon, line 21: again, here you should consider consistency, pick `std::string` or `const char*`

include/ofx-props.yml
```diff
@@ -0,0 +1,1508 @@
+########################################################################
```
abizeauMaxon, line 1: Is this auto generated? If so maybe add it into the header

abizeauMaxon (commented):
That super! I see a lot of good thing here! Will be helpfull with time.

include/ofxPropsMetadata.h
garyo: Thanks @abizeauMaxon -- totally agreed. As per discussion at the TSC today, I've moved these into `Support/include` and they are 100% C++. I used fixed-size `std::array`s where possible, as well as some other updates and cleanups. More feedback will be helpful; let me know if you try it out.

include/ofxPropsMetadata.h
garyo: I've tried now to move to `char *`, to avoid allocations where possible during startup. I should move to `std::string_view` in some places. I'll do that.

include/ofx-props.yml
garyo: `ofx-props.yml` is hand-created -- it's now the source of truth for all info about properties and property sets. That's the lion's share of the work in this PR (and probably the source of remaining bugs, where I may have gotten some stuff wrong). None of this info was machine-readable before, so I extracted it from the docs and header file comments.

garyo:
Superseded by #233, which landed the property metadata system in stages: the YAML property definitions, the generated C++ accessors, and docs (merged 2026-04-02). That delivers the property layer first, as @valadaptive requested here, while keeping params/images/clips for follow-up work. Closing this WIP PR in favor of #233; further refinement (e.g. inline `@propdef` blocks in the headers) has continued on main since then.

umajho:
> > I assume you'd like to see just the property layer landed soon, rather than waiting for more stuff like params, images, clips, etc.?
> 
> Yes. If it's possible to land this in stages rather than waiting for a complete overhaul, I think that would be good.
> 
> I might open an issue going over what would help for Rust bindings. I'm still experimenting and trying to come up with the best API.

Wow, I didn't realize that you were already considering writing Rust bindings for OpenFX.

I actually discovered OpenFX through your project ([ntsc-rs]), through your contributions to [egui]. Your work is really impressive!

Without knowing about your plans, I started working on my own Rust bindings here: https://github.com/kreantio/openfx-rs
(NOTE(2026/09/17): It is still in very early stages. The low-level bindings are about 80% complete, and I haven't started working on an idiomatic (according to me) Rust bindings/wrapper yet.)

In fact, [the code for ntsc-rs' OpenFX plugin](https://github.com/ntsc-rs/ntsc-rs/tree/main/crates/openfx-plugin) was my starting point for learning how to write OpenFX plugins (in Rust) and how to bundle them.

Thank you so much!

[ntsc-rs]: https://github.com/ntsc-rs/ntsc-rs
[egui]: https://github.com/emilk/egui

PR #169, opened 2024-09-06
Rename colour management style defs

### LINKED ISSUE #168: Incorrect value names in colour header
### Problem

`ofxColour.h` includes the following definitions which should not contain the word `Prop`, because they are not properties:

```kOfxImageEffectPropColourManagementNone
kOfxImageEffectPropColourManagementBasic
kOfxImageEffectPropColourManagementCore
kOfxImageEffectPropColourManagementFull
kOfxImageEffectPropColourManagementOCIO```

john-paulsmith:
The colour management style definitions contained the string ‘Prop’ which was incorrect because they aren’t properties. This complicated the introspection project so has been corrected. Fixes #168.


garyo (approved):
LGTM; since this was so recently added I think we can pretty safely make these changes.

PR #174, opened 2024-10-25
Fix doxygen reference for OfxTimeLineSuiteV1

valadaptive:
The documentation page for `OfxTimeLineSuiteV1` was incorrectly referencing `OfxImageEffectSuiteV1` instead.


garyo (approved):
LGTM -- thanks!

PR #178, opened 2024-12-19
Update colourspace example for more complete coverage

barretpj:
Example now includes 3 plugins, to exercise Basic, Core and Full styles.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -396,20 +396,29 @@ jobs:
           tar -czf openfx-$RELEASE_NAME.tar.gz -C Install OpenFX
 
       - name: Sign header/libs tarball with Sigstore
+        if: github.event_name == 'release'
         uses: sigstore/gh-action-sigstore-python@f514d46b907ebcd5bedc05145c03b69c1edd8b46 # v3.0.0
-        # if: github.event_name == 'release'
         with:
           inputs: openfx-${{ env.RELEASE_NAME }}.tar.gz
           upload-signing-artifacts: false
           release-signing-artifacts: false
 
       - name: Upload header/libs tarball and signatures
+        if: github.event_name == 'release'
         uses: actions/upload-artifact@v4
         with:
           name: "openfx-${{ env.RELEASE_NAME }}"
           path: |
             openfx-${{ env.RELEASE_NAME }}.tar.gz
             openfx-${{ env.RELEASE_NAME }}.tar.gz.sigstore.json
+
+      - name: Upload header/libs tarball (no signatures)
+        if: github.event_name != 'release'
+        uses: actions/upload-artifact@v4
+        with:
+          name: "openfx-${{ env.RELEASE_NAME }}"
+          path: |
+            openfx-${{ env.RELEASE_NAME }}.tar.gz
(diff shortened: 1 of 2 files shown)


revisionfx:
Thanks

Is there a difference between?:
  ROLE(kOfxColourspaceRoleColorTiming, "Color Timing role (Arbitrary scene log)"),
  ROLE(kOfxColourspaceRoleCompositingLog, "Compositing Log role (Arbitrary scene log)"),
Is there a difference between?:
  ROLE(kOfxColourspaceRoleMattePaint, "Matte Paint role"),
  ROLE(kOfxColourspaceRoleTexturePaint, "Texture Paint role (Typically sRGB)"),

barretpj:
> Is there a difference between?: ROLE(kOfxColourspaceRoleColorTiming, "Color Timing role (Arbitrary scene log)"), ROLE(kOfxColourspaceRoleCompositingLog, "Compositing Log role (Arbitrary scene log)"), Is there a difference between?: ROLE(kOfxColourspaceRoleMattePaint, "Matte Paint role"), ROLE(kOfxColourspaceRoleTexturePaint, "Texture Paint role (Typically sRGB)"),

Not relevant to this PR. These are the spaces/roles define in the OCIO config we use.

Guido-assim:
These plugins are very nice to check the behavior of the host :-).
Maybe a hypothetical case, but if the plugin specifies a basic color space as an output color space then the host doesn't know what the exact color space is that is used. The host might need to convert the output of the plugin to another color space for display or other reasons.

john-paulsmith:
> Maybe a hypothetical case, but if the plugin specifies a basic color space as an output color space then the host doesn't know what the exact color space is that is used. The host might need to convert the output of the plugin to another color space for display or other reasons.

Not specifically relevant to this PR, but my recommendations would be:

1. Plug-ins should use cross-referencing for the output clip unless absolutely necessary.
2. When using the basic style, plug-ins should not perform colourspace transformations to anything other than `ofx_raw` (e.g. motion vectors).
3. If the plug-in _does_ set a basic colourspace on the output clip, the host should interpret that in the same way it would if it was supplying images to the plug-in using that basic colourspace.

garyo:
Good, CI is passing again now. To fix the DCO check problem, see here https://github.com/AcademySoftwareFoundation/openfx/pull/178/checks and make sure to always sign commits with -s .

To make sure I never forget, I do this:
```
git config --global commit.template ~/.gitmessage
```
And ensure `Signed-off-by: Your Name <your.email@example.com>` is included in that template (I put it after two blank lines).

barretpj:
OK that eventually worked. I really struggle to get my head around this workflow.

garyo-gpsw:
Thanks for sticking with it, Phil! The signoff is just to make sure we are legally protected; sorry it's a pain.

garyo:
Thanks for sticking with it, Phil! The signoff is just to make sure we are legally protected; sorry it's a pain.

garyo:
@john-paulsmith do you approve this PR?

Guido-assim (approved):
Looks good to me. Very useful to test the behavior of a host.

PR #187, opened 2025-05-06
Add `OfxImageEffectPropNoSpatialAwareness`

### LINKED ISSUE #127: Add a "LUT" render argument
## Summary

Users often want to create a LUT from their image pipeline.

DaVinci Resolve achieves this by piping a LUT image through the node graph. They've defined an extension `OfxImageEffectPropNoSpatialAwareness` that allows the plugin to specify whether or not it wants to participate in the LUT generation process. This approach is smart, but I think there could be better logic.

It seems like there should be three options for communicating this to the host:

1. The plugin will *never* affect spatial characteristics of the image, and a LUT can be generated at any time.
2. The plugin will *always* affect spatial characteristics of the image, and should not be used in the LUT generation process. (Probably should be the default if not specified?).
3. The plugin might *currently* affect spatial characteristics of the image, but when requested during a render (perhaps a property on the source clip) can temporarily *disable those spatial characteristics* to help generate a LUT. (imagine a plugin that applies film grain, blurs, any sort of geometric distortion, etc. but also non-spatial color characteristics)

It might also be useful to provide helper code for the host for generating a LUT image in-memory to pipe through their image pipeline.

## Motivation
Users of hosts want the ability to generate LUTs. Hosts should have some built-in ability to do so. Plugins should be able to opt-in or opt-out.

## Impact
Purely additive.

## Documentation Impact
We'd need to add documentation, yes!

## Stakeholders
Plugins and Hosts!

gregcotten:
Plugins should be able to opt-in to host's LUT generators (or other non-spatial renders), and hosts need a way of indicating to the plugin that the current render requires rendering without spatial awareness. 

Resolve has already defined an extension `OfxImageEffectPropNoSpatialAwareness`, so I propose using that for both the plugin descriptor and the read-only render argument passed by the host.

Closes #127

### DIFF
--- a/include/ofxImageEffect.h
+++ b/include/ofxImageEffect.h
@@ -1175,15 +1175,16 @@ Also note that some hosts do implement kOfxImageEffectPropRenderScale - these tw
 #define kOfxImageEffectPropRenderQualityDraft "OfxImageEffectPropRenderQualityDraft"
 
 /** @brief Indicates that the plugin can render without spatial awareness, either inherently or by
-disabling certain parameters at render time. If the plugin descriptor has this prop set to "true", the 
-plugin is expected to disable spatial effects when the host sets this prop to "true" in the render call arguments.
+disabling certain parameters at render time. 
+
+If the plugin descriptor has this property set to "true", the plugin is expected to disable spatial effects when the host sets this property to "true" in the arguments passed to kOfxImageEffectActionBeginSequenceRender and kOfxImageEffectActionRender.
 
     - Type - string X 1
     - Property Set - plugin descriptor (read/write), render calls - host (read-only)
     - Default - "false"
     - Valid Values - This must be one of
-      - "false"  - the plugin cannot render without spatial awareness and the host will bypass it for renders that require no spatial awareness. 
-      - "true"   - the plugin can render without spatial awareness, and the host will indicate this type of render by setting kOfxImageEffectPropNoSpatialAwareness to "true" in the render call.
+      - "false"  - the plugin cannot render without spatial awareness and the host should bypass it for renders that require no spatial awareness.
+      - "true"   - the plugin can render without spatial awareness. The host will indicate this type of render by setting kOfxImageEffectPropNoSpatialAwareness to "true" in the arguments passed to kOfxImageEffectActionBeginSequenceRender and kOfxImageEffectActionRender.
  */
 #define kOfxImageEffectPropNoSpatialAwareness "OfxImageEffectPropNoSpatialAwareness"
 
--- a/include/ofxImageEffect.h
+++ b/include/ofxImageEffect.h
@@ -407,6 +407,7 @@ These are the list of actions passed to an image effect plugin's main function. 
      -  \ref kOfxImageEffectPropSequentialRenderStatus whether the effect is currently being rendered in strict frame order on a single instance
      -  \ref kOfxImageEffectPropInteractiveRenderStatus if the render is in response to a user modifying the effect in an interactive session
      -  \ref kOfxImageEffectPropRenderQualityDraft if the render should be done in draft mode (e.g. for faster scrubbing)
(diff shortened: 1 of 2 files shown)


garyo (approved):
LGTM! Once this is merged, please add a note in `release-notes.md` for it.

include/ofxImageEffect.h
```diff
@@ -1173,6 +1174,19 @@ Also note that some hosts do implement kOfxImageEffectPropRenderScale - these tw
  */
 #define kOfxImageEffectPropRenderQualityDraft "OfxImageEffectPropRenderQualityDraft"
 
+/** @brief Indicates that the plugin can render without spatial awareness, either inherently or by
+disabling certain parameters at render time. If the plugin descriptor has this prop set to "true", the
```
barretpj, line 1178: property, not prop, please

include/ofxImageEffect.h
```diff
@@ -1173,6 +1174,19 @@ Also note that some hosts do implement kOfxImageEffectPropRenderScale - these tw
  */
 #define kOfxImageEffectPropRenderQualityDraft "OfxImageEffectPropRenderQualityDraft"
 
+/** @brief Indicates that the plugin can render without spatial awareness, either inherently or by
+disabling certain parameters at render time. If the plugin descriptor has this prop set to "true", the 
+plugin is expected to disable spatial effects when the host sets this prop to "true" in the render call arguments.
```
barretpj, line 1179: "in the arguments passed to kOfxImageEffectActionBeginSequenceRender and kOfxImageEffectActionRender" would be clearer

gregcotten:
OK, I'm all set. Just waiting to see if @garyo would like the change to `release-notes.md` added to this PR

garyo:
If you don't mind rebasing this onto latest main, where I just started a 1.6 section of release-notes.md, then yes please, that would be great. Rebase and add to the relnotes.

gregcotten:
> If you don't mind rebasing this onto latest main, where I just started a 1.6 section of release-notes.md, then yes please, that would be great. Rebase and add to the relnotes.

done

revisionfx:
I saw a comment about kOfxImageEffectPropRenderQualityDraft 
What would be more useful to us is kOfxImageEffectPropRenderQualityThumbnails 

Pierre

gregcotten:
> I saw a comment about kOfxImageEffectPropRenderQualityDraft What would be more useful to us is kOfxImageEffectPropRenderQualityThumbnails

Might want to make a new issue about that. This one is closed!

barretpj:
> > I saw a comment about kOfxImageEffectPropRenderQualityDraft What would be more useful to us is kOfxImageEffectPropRenderQualityThumbnails
> 
> Might want to make a new issue about that. This one is closed!

Done - #193

barretpj:
(It should really be NoTemporal as well...)

revisionfx:
DO you mean multi-frame processing (get 3 frames from input) - Yes|OK - But what things like basic time offsets performed by plugin (even if only a single frame is requested)?..  Some hosts have a single thumbnail (a one time thing using first or current frame) while other populates on a timeline multiple frames.

gregcotten:
I believe @barretpj is referring to a render with NoSpatialAwareness set to "true" should also mean it has "no temporal awareness"

gregcotten:
> This is nothing to do with thumbnails. This is when we want to do a render which **only** changes the colour of pixels based on their input colour, so the host can analyse the overall render and approximate the changes with a LUT. If the plugin is considering other pixels, or pixels from images at other times, then the image analysis in the host will most likely fail.
> 
> 

Sensible. We can adjust the description accordingly.

barretpj:
@revisionfx Plugins must only set NoSpatialAwareness="true" if they are able to operate in a change-current-frame-pixel-colour-only mode (either all the time, or when requested by the host via NoSpatialAwareness="true" in the render args)

PR #191, opened 2025-05-06
Require hosts to call instanceChanged action after loading project

### LINKED ISSUE #184: Require hosts to call Instance Changed action after loading a project
# Open Effects Proposal for Standard Change

Please read the contribution [guidelines](https://github.com/ofxa/openfx/wiki/Extending-OpenFX-Guidelines#submit-a-proposal-or-bug-report-to-review) first.

## Standard Change Workflow
- [x] Create proposal as issue (you're doing this now!)
- [x] Tag this issue with `standard change` tag
- [ ] Identify subcommittee: at least one plug-in vendor, and at least one host
- [x] Discuss the idea in this issue
- [ ] Write new or updated code and doc
- [ ] Publish updates as a pull request (ideally on a `feature/PROPOSAL-NAME` branch)
   - [ ] Make sure that PR references this issue number to keep them in sync
   - [ ] Discuss and review code in the PR
   - [ ] Meet all requirements below for accepting PR
- [ ] When subcommittee signs off and other members don't have any further review comments, 
      maintainer merges PR to master which closes PR and issue

## Requirements for accepting a standard change:
- [ ] Header files updated
- [ ] Documentation updated
- [ ] Release notes added
- [ ] Compatibility review completed
- [ ] Working code demonstrated with at least one host and one plugin
- [ ] At least two members sign off
- [ ] No further changes requested from membership

## Summary

Currently a host may or may not call instanceChanged after loading a saved project. This means plugins do not get an opportunity to update the enabled status of params or take other actions based on the input clips and/or param values. This proposal would *require* conforming hosts to call `kOfxActionInstanceChanged` after loading a project containing OpenFX plugin instances. This call would be with type=`kOfxTypeClip`. `kOfxTypeParameter` would also make sense but I suggest we just use Clip, and plugins can do whatever they need to. 

As for `kOfxPropChangeReason`, should we recomend `kOfxChangeUserEdited` since loading a project is a user-initiated action? We could add a new reason, `kOfxChangeProjectLoaded` if that is needed, to distinguish these events from others, but since some hosts are already doing this, I don't think we need to make them change their behavior.

## Motivation

Without this, plugins can't set their param enabledness or some overlay features because nothing is called after createInstance. A plugin has no way to know it's been updated (clip connected, params modified) after the instance was created.

## Problem

This would enable plugins to show proper param status after a project is loaded.

## Impact

This should have no impact on existing plugins, since some hosts already call instanceChanged after loading a project, and a plugin should respond to instanceChanged whenever it happens. 

It does impact hosts, because any conforming host will need to make this call, if it doesn't already.

## Documentation Impact

This is primarily (or entirely) a documentation/standard change. No change to the existing API.

## Stakeholders

Both plugins and hosts will benefit from this by improving UI behavior.

garyo:
Fixes #184.


Guido-assim (commented):
Would reason kOfxChangeUserEdited not make more sense for instance changed? It is the user/host who attached the clips not the plugin itself.

PR #192, opened 2025-05-06
Add Windows ARM64 plugin install locations with rationale

### LINKED ISSUE #160: Windows ARM64 and ARM64EC support
# Open Effects Proposal for Standard Change

Please read the contribution [guidelines](https://github.com/ofxa/openfx/wiki/Extending-OpenFX-Guidelines#submit-a-proposal-or-bug-report-to-review) first.

## Standard Change Workflow
- [x] Create proposal as issue (you're doing this now!)
- [x] Tag this issue with `standard change` tag
- [ ] Identify subcommittee: at least one plug-in vendor, and at least one host
- [x] Discuss the idea in this issue
- [ ] Write new or updated code and doc
- [ ] Publish updates as a pull request (ideally on a `feature/PROPOSAL-NAME` branch)
   - [ ] Make sure that PR references this issue number to keep them in sync
   - [ ] Discuss and review code in the PR
   - [ ] Meet all requirements below for accepting PR
- [ ] When subcommittee signs off and other members don't have any further review comments, 
      maintainer merges PR to master which closes PR and issue

## Requirements for accepting a standard change:
- [ ] Header files updated
- [ ] Documentation updated
- [ ] Release notes added
- [ ] Compatibility review completed
- [ ] Working code demonstrated with at least one host and one plugin
- [ ] At least two members sign off
- [ ] No further changes requested from membership

## Summary

OpenFX plugins are installed in standard locations, based on the ABI architecture. See https://openfx.readthedocs.io/en/main/Reference/ofxPackaging.html#installation-directory-hierarchy for the current list. 

On Windows, the spec currently says:

> On Windows with arm64, `Win-arm64` should be used (vs. current `Win32` and `Win64` which are Intel-specific)

Windows 11 will soon be supporting two new architectures, `arm64` and `arm64ec`. No machines currently support this, but such machines are expected in summer of 2024. According to [Microsoft](https://learn.microsoft.com/en-us/windows/arm/arm64ec), `arm64ec` supports running Intel code in emulation, within an arm64 application on an arm64 CPU. This should allow Intel-based plugins to run in an arm64ec host. Native `arm64` applications will presumably fail to load an Intel dll, so the spec should be specific about where plugins for each architecture are installed.

We will need to confirm once machines and host apps are available, but I believe this will work:
* An Arm64EC host should attempt to load plugins from `.../Win-arm64` (which will be arm64 native) and then from `.../Win64` (which will be Intel). 
* An Arm64 native host should load plugins from `.../Win-arm64`.
* We do not need to provide a `Win-arm64ec` plugin install dir. All arm64 plugins, whether built as arm64 or arm64ec, should be installed in `Win-arm64`.

## Impact

This will impact all hosts and plugins which want to support Arm64 on Windows.

## Documentation Impact

The only change to the doc (the spec) is to update the install dirs list, and add rationale as per the discussion in this issue.

garyo:
Fixes #160, which see.

### DIFF
--- a/Documentation/sources/Reference/ofxPackaging.rst
+++ b/Documentation/sources/Reference/ofxPackaging.rst
@@ -123,15 +123,16 @@ cannot load native Arm64 DLLs, only Intel or Arm64EC.
 Windows also supports an Arm64X "fat" binary PE format which
 contains both Intel/Arm64EC and Arm64 binaries.
 
-Since apps built natively for Arm64 ("Arm64" abi) can only load Arm64
-DLLs, those apps should attempt to load plugins from the "Win-arm64x"
-folder first, then "Win-arm64". Arm64EC hosts (running on an Arm64
-system) should try the "Win-arm64x" folder first, then "Win-arm64ec".
-Arm64EC hosts can also use x64 DLLs, so they may try "Win64" as well,
-but should prefer the other folders to avoid running the plugin in
-emulation. An Arm64X host will be running either its Arm64EC branch or
-its Arm64 branch depending on how it's launched, so it should load
-plugins according to the above depending on the current mode.
+**Search Order**: Since apps built natively for Arm64 ("Arm64" abi)
+can only load Arm64 DLLs, those apps should attempt to load plugins
+from the "Win-arm64" folder first, then "Win-arm64x". Arm64EC hosts
+(running on an Arm64 system) should try the "Win-arm64ec" folder
+first, then "Win-arm64x". Arm64EC hosts can also use x64 DLLs, so they
+may try "Win64" as well, but should prefer the other folders to avoid
+running the plugin in emulation. An Arm64X host will be running either
+its Arm64EC branch or its Arm64 branch depending on how it's launched,
+so it should load plugins according to the above depending on the
+current mode.
 
 Note that there is no "fat binary" format on Windows that contains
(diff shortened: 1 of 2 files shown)


john-paulsmith:
On reflection, I think the search logic should be reversed. We don't yet know if developers will find limitations or performance degradation using the Arm64X wrapper, so it makes sense to search the native location first.

garyo:
> On reflection, I think the search logic should be reversed. We don't yet know if developers will find limitations or performance degradation using the Arm64X wrapper, so it makes sense to search the native location first.

Good point. It does make sense to look in the "most native" folder first. I'll make that update.

garyo:
> On reflection, I think the search logic should be reversed. We don't yet know if developers will find limitations or performance degradation using the Arm64X wrapper, so it makes sense to search the native location first.

Done.

revisionfx:
I can confirm that Resolve WIndows Arm is EC (just downloaded). WIll try this week what happens if we download our AMD64 plugins on that machine :)

Meaning we will likely at some point need to have two distinct builds anyway on Windows Arm. Arggh! 
Then seems it would be more practical to just have ARM64X folder which supports all - it's just a bundle - a one time thing for app to choose the right dll to load in a single folder.  I say that before I see what happens to GPU support in different modes... We also need ARM64 for linux.

Issue #203, opened 2025-06-16
Effect Instances should have a unique identifier


gregcotten:
Stubbing this for now, but the short of it:

It would be very useful for the host to give an effect instance some sort of unique identifier (probably an actual UUID string) that can be used to seed "unique to this instance" behavior, such as seeding a random number generator. This identifier shouldn't be copied if the user copies the plugin to another shot, for example. That would have its own new unique identifier. The identifier should persist between project loads/unloads, and not change if a plugin instance is deleted and then undo-deleted, for example.

I think some folks will ask: "What defines a unique instance of an effect?". My answer is that you define it however it's defined to your end-users. Most hosts probably have a concrete answer for this and I would imagine it's all pretty similar, but please do make edge and corner cases known!

barretpj:
So to try to pin down the language you mean each "instance" of the OFX effect that the user can see in their project/timeline/graph/etc is given (by the host) a unique ID, and this ID is set as a property in every OFX plugin effect instance created by/from/for that "instance" (and there could be many of these, especially in a host which uses ephemeral render instances).

gregcotten:
> So to try to pin down the language you mean each "instance" of the OFX effect that the user can see in their project/timeline/graph/etc is given (by the host) a unique ID, and this ID is set as a property in every OFX plugin effect instance created by/from/for that "instance" (and there could be many of these, especially in a host which uses ephemeral render instances).

Yes!

gregcotten:
> My take is render instances for the same user-applied instance would all have the same ID. If it's used as a seed, then it would need to be consistent across render instances.

I believe that is what @barretpj is saying!

barretpj:
> If you need temporally coherent randomness in your render you cannot seed your RNG based on an Image UID. With a Clip UID (or Effect UID) you seed off of that UID and then you can use the time property to be temporally coherent across the Clip or Effect instance respectively.

Are you talking about applying different plugin operators at adjacent points to the same upstream clip, and using a clip UID to make them coherent?

> I'd settle for just an Effect UID and can update my PR accordingly if you think the Clip UID won't work. Or we can narrow the scope of the Clip UID property to only support Source clips if you think that would help.

Unclear what you mean by "only support Source clips" - in the General context there are no rules about what your plugin names its inputs - see https://openfx.readthedocs.io/en/main/Reference/ofxImageEffectContexts.html

gregcotten:
[...]

PR #204, opened 2025-06-18
Add a Unique Identifier property to ImageEffect instances

gregcotten:
As discussed in #203


fxtech:
@gregcotten are you able to check the UUID stuff in Silhouette 2025.5 or 2026?

fxtech:
NOTE: after discussing this, we need to update this PR with documentation explaining the limitations of this feature if plugins want to use it as a *seed*.

gregcotten:
@fxtech, I've just installed and activated a trial for Silhouette 2026.0.2 but cannot figure out how to get it to load a OpenFX plugin...

fxtech:
> @fxtech, I've just installed and activated a trial for Silhouette 2026.0.2 but cannot figure out how to get it to load a OpenFX plugin...

Hey @gregcotten sorry I didn't see that!
In Silhouette 2026+ you need to whitelist the OFX plugins you want it to load. Go to Preferences->Plugins.

BUT - we agreed in the OFX Meeting that this shouldn't be used as a seed, due to the way timeline-based hosts work.
There MIGHT be a use for this in some cases but it shouldn't be used as a seed.

I have an alternate implementation in mind for seed use, which would be a new property on Int parameters that would that could provide that capability.

Issue #206, opened 2025-07-25
Change copyright notice to match ASWF standards


jmertic:
Following up from https://github.com/AcademySoftwareFoundation/openfx/pull/110 and https://github.com/AcademySoftwareFoundation/openfx/pull/205#discussion_r2231933082, we should change the copyright notice to:

```
Copyright Contributors to the OpenFX project
```

garyo:
Note that in #110 it was noted that "according to the [legal docs](https://www.dropbox.com/home/OFXA/Contracts?preview=IP-transfer-to-ASWF-2022.pdf) it's "OpenFX a Series of LF Projects, LLC", and I'm assuming it's reasonable to abbreviate that to OpenFX." so this change would remove the attribution of copyright to the Series itself (which owns the copyright -- that is part of the IP transfer legal document; copyrights were all transferred from the Association to the Series). I'm not sure the legal ramifications of that.

lilith887:
Hi, I saw the discussion about updating the copyright notice. I'd like to help with the modification if possible.

garyo:
Hi @lilith887; we would be happy to review a PR changing all instances of `Copyright OpenFX and contributors to the OpenFX project` to `Copyright Contributors to the OpenFX project`. Just a note that before we can merge the PR we'll need a small legal review though. Thanks for your offer and your support for OpenFX!

jmertic:
If the TSC believes it is crucial to highlight the distinction, there are no issues with the current copyright notice continuing. It could be helpful to call out the provenance of the original codebase, as you mentioned, just for future reference.

lilith887:
Hi team, I've reverted all copyright notice changes as discussed in the TSC meeting, restoring the original notices for all files. Please take a look at the updated PR when you have time. Thanks

PR #209, opened 2025-08-13
Correct copyright notice, with explanation

john-paulsmith:
Questions have arisen over the OpenFX copyright notice on more than one occasion (#110, #206). CONTRIBUTING.md now includes the correct copyright notice and explanation of why it is used.


garyo (approved):
LGTM—thanks for doing this!

PR #216, opened 2025-09-27
Fix typos in Examples subdirectory

luzpaz:
Found via codespell


Examples/ColourSpace/colourspace.cpp
```diff
@@ -543,7 +543,7 @@ getClipPreferences( OfxImageEffectHandle effect, OfxPropertySetHandle /*inArgs*/
   OfxPropertySetHandle effectProps;
   gEffectHost->getPropertySet(effect, &effectProps);
 
-  // retieve the colour management style that the host has decided we shall use
+  // reteive the colour management style that the host has decided we shall use
```
garyo, line 546: Should be "retrieve" (and below)

Issue #217, opened 2025-10-07
Conform Effects


john-paulsmith:
# Open Effects Proposal for Standard Change

Please read the contribution [guidelines](https://github.com/ofxa/openfx/wiki/Extending-OpenFX-Guidelines#submit-a-proposal-or-bug-report-to-review) first.

## Standard Change Workflow
- [x] Create proposal as issue (you're doing this now!)
- [x] Tag this issue with `standard change` tag
- [x] Identify subcommittee: at least one plug-in vendor, and at least one host
- [ ] Discuss the idea in this issue
- [ ] Write new or updated code and doc
- [ ] Publish updates as a pull request (ideally on a `feature/PROPOSAL-NAME` branch)
   - [ ] Make sure that PR references this issue number to keep them in sync
   - [ ] Discuss and review code in the PR
   - [ ] Meet all requirements below for accepting PR
- [ ] When subcommittee signs off and other members don't have any further review comments, 
      maintainer merges PR to master which closes PR and issue

## Requirements for accepting a standard change:
- [ ] Header files updated
- [ ] Documentation updated
- [ ] Release notes added
- [ ] Compatibility review completed
- [ ] Working code demonstrated with at least one host and one plugin
- [ ] At least two members sign off
- [ ] No further changes requested from membership

## Summary

Users want to move a project from a _foreign_ (non-OpenFX) host to an OpenFX host. This proposal includes APIs to facilitate the process.

## Motivation

Most shows have a two-stage editorial pipeline. A variety of names are used for these stages, but I will use the terms _offline_, where creative editorial decisions are made, and _online_, which includes colour grading, final VFX and other finishing tasks. The process of transferring projects from offline to online is _conform_, i.e. _conforming_.

Conform is facilitated using an EDL (edit decision list) file, which is exported from the offline host and imported into the online host. The dominant format is AAF, although OpenTimelineIO starting to become more popular. The EDL format used isn't important for this proposal, as long as it contains complete information about effects used in the offline.

This process often discards effects for a variety of technical reasons, and many users have requested we improve this situation so effects flow through the conform process as seamlessly as possible.

## Problem

[...]

PR #218, opened 2025-10-07
Feature/conform

john-paulsmith:
This change adds new properties and an action to help hosts with editorial timeline conform, as described in #217.

`kOfxPropAka` will allow plug-ins to list alternative names by which any named object could be known. A `kOfxPropAka2D` property covers multidimensional parameters. For example, an effect could list the name it is known by in a different API such as AVX or Adobe. It can also list the names its parameters are known by, which might be GUIDs or similar.

The conform action, `kOfxActionConform`, will allow plug-ins to update their parameter values and keyframes based on information passed from the host. Prior to invoking this action, hosts will set parameter values to the best of their ability, and the plug-in will examine these and update them based on internal knowledge about where they came from. For example, a spatial parameter might need to be adjusted to compensate for different origin.

### DIFF
--- a/include/ofxCore.h
+++ b/include/ofxCore.h
@@ -669,6 +669,15 @@ This property is used to label objects uniquely amoung objects of that type. It 
 */
 #define kOfxPropName "OfxPropName"
 
+/** @brief Alternative names for an object.
+
+    - Type - string X N
+    - Property Set - on many objects (descriptors and instances), see \ref PropertiesByObject (read only)
+
+This property is used to provide alternative names for objects. It can be used in any situation where ::kOfxPropName is used, and is helpful for situations where the effects are being transferred from another plug-in API.
+*/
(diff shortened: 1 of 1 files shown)


include/ofxCore.h
```diff
@@ -705,6 +705,15 @@ This property is used to provide alternative names for objects. It can be used i
 */
 #define kOfxPropAka "OfxPropAka"
 
+/** @brief Alternative names for parameter dimensions.
+
+    - Type - string X N X M
```
barretpj, line 710: This is, I assume, for the case where the OFX plugin uses a single parameter with a 2D param type, but one foreign host uses (e.g.) `TranslateX` and `TranslateY` and another uses `tx` and `ty`. How do you store those mappings in a 1D string array property? What does "string X N X M" mean?

PR #224, opened 2025-11-14
Fix Windows #defines—don't define WINDOWS or _WINDOWS, use predefined _WIN32 everywhere

garyo:
There's no need to have our own `WINDOWS` #define; Windows always predefines `_WIN32`. Tested with our CMake/Conan build on Mac & Windows. This should clean things up in preparation for publishing a proper Conancenter recipe for OpenFX that doesn't cause users of the recipe to get WINDOWS definitions in their builds.


jfpanisset (approved):
Looks good, avoiding custom symbols that look like system defined ones is always good.

PR #225, opened 2025-11-14
Add release notes for upcoming 1.5.1 release

garyo:
This needs to go out soon so I can update the Conancenter repo with Windows fixes (see also #224 , and please review that as well so I can merge it), and check out https://github.com/conan-io/conan-center-index/pull/26667 if you're interested in why the Windows stuff matters in the conanfile.

### DIFF
--- a/release-notes.md
+++ b/release-notes.md
@@ -29,7 +29,7 @@ This is version 1.5.1 of the OpenFX API, a minor release with bug fixes and impr
 
 ## Key Features of OpenFX Version 1.5.1:
 
-- **CPU Render Support**: Added `kOfxImageEffectPropCPURenderSupported` property to allow plugins to declare CPU rendering capability.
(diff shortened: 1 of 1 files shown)


release-notes.md
```diff
@@ -19,14 +19,48 @@ This is version 1.6 of the OpenFX API.
 
 ## Fixes in OpenFX Version 1.6:
```
barretpj: Should this still say 1.6 (here and above)?

barretpj (changes requested):
Feedback added

garyo:
Thanks for the comments, Phil! I addressed them above. I deliberately left the 1.6 section above, to serve as a place to collect things that will go into 1.6 (that's how we usually do it, though I agree it can be confusing—suggestions appreciated! I moved the SpatialAwareness prop into 1.5.1, which is why it looked like it was getting deleted.

release-notes.md
garyo: We've always had a placeholder for the next release... but we could keep "un-released" changes elsewhere if you think that would help. It would certainly make this cleaner.

PR #232, opened 2026-03-03
Add #ifndef guard for NOMINMAX macro define

kevreco:
Down the line some projects consuming openfx source code defines there own `NOMINMAX` define via command line.

This result in few warnings like:
```
warning C4005: 'NOMINMAX': macro redefinition
```

This MR fixes that.


garyo:
This makes sense. Better would be to undef it after including windows.h though, to avoid leaking that define into the user's code. Could you update this PR for that? Also you'll need to sign off on it per our DCO requirements. You can just force-push an updated commit to this PR.

PR #233, opened 2026-03-03
Add property metadata system: YAML definitions, generated C++ accesso…

garyo:
Add property metadata system: YAML definitions, generated C++ accessors, and docs
    
    Introduce ofx-props.yml as the structured source of truth for all OpenFX
    property definitions. Add generator scripts (gen-props.py, gen-props-doc.py)
    that produce C++ metadata headers (ofxPropsMetadata.h, ofxPropsBySet.h) and
    type-safe property set accessor classes (ofxPropSetAccessors.h,
    ofxPropSetAccessorsHost.h) for both plugins and hosts. Also generates
    structured Sphinx documentation for the properties reference.
    
    Includes fixes to property type definitions in C headers (e.g. RoD/RoI
    types in ofxImageEffect.h) and a host-specific property extension example.


garyo:
@john-paulsmith thanks for your push in the right direction — the YAML file is gone now, and all properties are machine-readably documented right in the header, using @propdef / @actiondef sections. There's doc in Documentation/README.md. Should be easier to maintain. The readthedocs docs should be nicer now too.

Reviewers: I think this is worth at least a scan of the whole changeset. You may see things that your host disagrees with; please call those out if you see them.

Guido-assim (commented):
I see a few cases of "dimension: 0" in a propdef, I wonder if that is valid?

garyo:
> I see a few cases of "dimension: 0" in a propdef, I wonder if that is valid?

That means "any", or unspecified. For params, for instance, it's the number of dims in the param. For PropVersion, it's whatever the plugin wants.

Guido-assim:
> > I see a few cases of "dimension: 0" in a propdef, I wonder if that is valid?
> 
> That means "any", or unspecified. For params, for instance, it's the number of dims in the param. For PropVersion, it's whatever the plugin wants.

Is that different from 'N' that is mentioned in README.md, or has 'N' been replaced by '0'?

garyo:
> Is that different from 'N' that is mentioned in README.md, or has 'N' been replaced by '0'?

Ah, good point -- I'll review that.

Documentation/sources/Reference/ofxPropertiesReferenceGenerated.rst
```diff
@@ -0,0 +1,2253 @@
+.. _propertiesReferenceGenerated:
+Properties Reference
+====================
+
+This reference is auto-generated from property definitions in the OpenFX source code.
+It provides a structured view of properties with their types, dimensions, and where they are used.
+For each property, a link to the detailed Doxygen documentation is provided when available.
+
+
+Boolean Properties
```
barretpj: OFX doesn't have a concept of boolean properties, except in the XML dtd which I don't think anyone really uses?

Documentation/sources/Reference/ofxPropertiesReferenceGenerated.rst
```diff
@@ -0,0 +1,2253 @@
...
+**OfxImageEffectPropThumbnailRender**
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- **C #define**: :c:macro:`kOfxImageEffectPropThumbnailRender`
+- **Type**: enum
+- **Dimension**: 1
+- **Valid Values**:
+  - ``false``
+  - ``true``
+- **Introduced in**: version 1.5.1
```
barretpj, line 1324: enum false/true is arguably more boolean than int 0/1 :-)

openfx-cpp/examples/host-specific-props/myhost/myhostPropsMetadata.h
```diff
@@ -0,0 +1,101 @@
...
+// Base template (leave undefined - specializations required)
+template<PropId id>
+struct PropTraits;
+
+template<>
+struct PropTraits<PropId::MyHostViewerProcess> {
+  using type = const char*;
+  static constexpr bool is_multitype = false;
+  static constexpr int dimension = 1;
+  static constexpr const openfx::PropDef& def = prop_defs[static_cast<int>(PropId::MyHostViewerProcess)];
```
barretpj: Pedantically these should be `static_cast<size_t>` or at least an unsigned type.

barretpj (commented):
Only done a very quick review pass. There's a LOT of code in here and as I don't use any of the C++ APIs I'm not best-placed to comment.

openfx-cpp/examples/host-specific-props/myhost/myhostPropsMetadata.h
revisionfx: Looks like we are loosing things like this?:
Property Set - interact instance property (read/write)

This is useful info

garyo: Fixed the cast. And thanks Pierre, I've restored the read/write info in the docs using the metadata.

Documentation/sources/Reference/ofxPropertiesReferenceGenerated.rst
garyo: Yes, and sadly we are not consistent on this. Just leaving this as is.

Issue #236, opened 2026-04-19
1.5.1: FMT_STRING() breaks spdlog in Examples/ColourSpace/colourspace.cpp


jfpanisset:
In 1.5.1 the calls to spdlog::info() wrap the string passed to spdlog in the FMT_STRING() macro.

On gcc 14.2.1 when spdlog() is called with a fstring-style string and variables, everything is fine:

```
spdlog::info(FMT_STRING("unknown colour management style '{}'"), style);
```

but when the string is a simple string without variable interpolation:

```
spdlog::info(FMT_STRING("Specifying preferred input colourspaces"));
```

the compiler rejects it, complaining about incomplete types.

I see this with spdlog 1.5.3 or 1.7.0, with either the corresponding fmt compiled separately, or using the one bundled in spdlog. Same issue if I patch CMakeLists.txt to force compiling in C++ 20 mode.

I am able to get around the issue with this gross patch:

```
// In some environments {fmt} needs FMT_STRING
//#ifndef FMT_STRING
#define FMT_STRING(x) x
//#endif
```

at the cost of gcc complaining about FMT_STRING being redifined.

I don't understand enough about spdlog, fmt or template metaprogramming to get at the root of the issue. I can provide more details on the errors reported by gcc if needed.

garyo:
I don 't see this with any recent fmt/spdlog with gcc, but in semi-recent fmt (e.g. with spdlog 1.8+), FMT_STRING is no longer needed, so I'll just remove it.

garyo:
PR merged. Closing.

PR #237, opened 2026-04-20
Fix issue #236: drop FMT_STRING wrappers in ColourSpace example

garyo:
spdlog since around 1.8 doesn't need FMT_STRING anymore, and it may cause type errors when used on plain strings with no format specs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,6 +81,27 @@ jobs:
             compiler-desc: Clang
             cuda: false
             opencl: true
+          - name_prefix: Linux Rocky 8 VFX CY2023 strict fmt
+            # Regression guard for openfx#236 / #237: ensures every
+            # fmt/spdlog callsite with format args uses OFX_FMT_STRING.
+            # Intentionally does not produce release artifacts.
+            release_prefix: linux-vfx2023-strict-fmt
+            ostype: linux
+            aswfdockerbuild: true
+            os: ubuntu-latest
+            container: aswf/ci-base:2023
+            vfx-cy: 2023
+            has_cmake_presets: false
+            buildtype: Release
+            conan_version: 2.20.1
+            cxx-standard: 17
+            cxx-compiler: clang++
+            cc-compiler: clang
+            compiler-desc: Clang
+            cuda: false
+            opencl: true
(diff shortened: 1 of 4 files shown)


barretpj:
It's needed for any compilation using C++17 or older, which includes my build environment.

garyo:
Interesting. I tried a bunch of compiler/OS/fmt/spdlog versions yesterday and couldn't make it fail without it, so please send the exact setup and I'll make sure that's tested and we'll add checks for it.

barretpj:
This is building the examples inside our Baselight makefile-based build, not using cmake or other build systems.

barretpj:
Specifically, gcc 11.5, fmt{} 11.1.4, `-std:c++17`

jfpanisset:
Here's what I ran into:

https://github.com/AcademySoftwareFoundation/openfx/issues/236

With g++ 14.2.1, compiling either C++17 or C++20, using spdlog 1.5.1 or 1.7.0, with either the embedded fmt version or the corresponding, external fmt, the compiler rejects calls to `spdlog::info(FMT_STRING("a simple string"))` without variable interpolation (it compiles fine when the string contains "{}" and extra variables are passed). The error complains about an incomplete type during template instantiation.

RIght now I'm having to override FMT_STRING():

```
#if 1
#define FMT_STRING(x) x
#endif
```

but of course the compiler complains. We could either try to figure out the specific compiler versions which have issues with this and #ifdef appropriately, or use our own OFX_FMT_STRING() macro so it can be easily #ifdef'd away without having to re-define the existing one?

garyo:
Thanks for these notes, folks — I'll look into this and come up with a proper fix.

jfpanisset (approved):
Looks like you actually went to the bottom of this!

BTW I'm about to release updated aswf-docker images, including an update to the 2023 images (which had not been updated in a long time since the transition of 2024/25/26 to Conan 2), I've verified that it builds against OpenFX 1.5s, but not latest. Will be good to test.

garyo:
Yeah, took a few tries (Claude helps a lot with this, spinning up containers, trying things quickly). But I think this is a complete solution. I wish that macro weren't needed at all, but for older compilers/fmt versions it looks like it is.

Issue #221, opened 2025-10-27
Is there a way to declare a plugin obsolete to the host app?


scritturamista:
Going through the headers and docs, I couldn't find a property to declare a plugin _obsolete_.

This is a common concept in other VFX SDKs. The purpose is to allow shipping a new version – which is obviously advertised in the UI – while bundling previous versions of the product for compatibility reasons.

To be clear: this is a boolean that tells the host app the plugin should be hidden in the UI for the purpose of adding new instances. The plugin should remain usable by the host app in order to support existing uses of the plugin, e.g.

```
/** @brief Indicates whether the plugin should be hidden in the UI while remaining fully functional where it had been previously used

   - Type - int X 1
   - Property Set - plugin descriptor (read/write)
   - Default - 0
   - Valid Values - This must be one of
       - 0 - which means the plugin is the current version, and should be advertised to the user
       - 1 -  which means the plugin should not be advertised to the user in the effect browser UI

*/
#define kOfxImageEffectPluginPropObsolete "OfxImageEffectPluginPropObsolete"
```

Did I miss anything?

Thanks!

barretpj:
I'm coming from the host vendor perspective, and I know that if a customer installs a plugin and it doesn't show up in the UI, they will report it to us as a bug in the host. And we will have to spend time diagnosing why. So I want the documentation in the header file to make it clear to all plugin developers what the best-practice recommended use of this flag is.

> I would imagine that the developer/company is in charge of avoiding this. If you ship an installer that places multiple versions of the same plugin on the system, it would seem natural that one of them (and only one of them) would be the non-obsolete one.

Yes, that's what I want to made explicit as in the comment documentation.

scritturamista:
@fxtech That is kind of you! I'll come back to this when ready. In my limited experience with OFX I have accumulated a long list of small feature requests, mostly gaps in the SDK that come from "diffing" with other SDKs I'm more familiar with. It seems the best approach is to create a branch with all these small changes and reuse that same branch for multiple PRs, each with its separate testing procedure but ultimately (and hopefully) with a single test plugin to work with.

garyo:
@scritturamista it's best for each PR to be separate, for us; that way each can be tested with plugins and hosts independently, we can discuss them and get them merged in cleanly.

scritturamista:
@garyo I should have been clearer: yes each change will be a separate PR, but is it OK if they are all off of main, or do you prefer every change on its own feature branch?

garyo:
[...]

PR #239, opened 2026-05-07
Add kOfxImageEffectPluginPropObsolete for issue #221

garyo:
Lets a plugin bundle ship newer and older versions of the same plugin together while hiding the obsolete versions from the host's effect browser. Obsolete plugins remain loadable so existing uses keep working; at least one version in the bundle must leave the property unset or 0 so it stays visible.

Header definition by Phil Barrett; generated metadata files (ofxPropsMetadata.h, ofxPropsBySet.h, ofxPropSetAccessors.h, ofxPropSetAccessorsHost.h) regenerated via scripts/gen-props.py.

See #221.


scritturamista:
@garyo @barretpj Thank you, extremely appreciated!

Issue #158, opened 2024-05-07
Color Managed Color Parameter


revisionfx:
## Standard Change Workflow
- [x ] Create proposal as issue (you're doing this now!)
- [x ] Tag this issue with `standard change` tag
 
## Summary

After we release colorspace suite, we should follow up with color managed color parameter.

## Example use cases:

set a red rectangle - user might think in terms of 1,0,0 but the actual given working process colorspace might not be intuitive.
pick somewhere on screen to get color somehow
set a threshold value slider
make your own custom widget (e.g. HLS)
map OS supplied color picker (some do that I think when you click on color param)

## Problem

likely requires a host call back to return to us chromacites?...
perhaps color are always define in linear internally? 
host might support different display options if they popup a color param editor for such param

john-paulsmith:
Now that we have colour management support in OFX, this issue becomes feasible to address. My proposed starting point would be to add a new string property `OfxParamPropColourspace` which stores the colourspace of the colour value stored in the param. The string would follow the same rules as `OfxImageClipPropColourspace`, i.e. it needs to be valid for the colour management style in use.

Some open questions:

1. Should this property be read/write for both host and plug-in? There are scenarios where the correct colourspace is known to the host (user picks from the image) and others where it is known to the plug-in (loading a preset or performing analysis of an input image).
2. If a plug-in loads a preset which was defined in a different colourspace to the one currently set on the parameter, would it actually be better from a user perspective for the plug-in to transform its colour values into the host's preferred space?

revisionfx:
Here's my notes from today


1. We have the advantage that the values are not displayed to user (is this true, does everyone display a color rectangle for color param?).

2. Assuming 1 - we could have 1 (one) color param space - which I think could just be linear AP0 ? What Rohit at BMD suggested a while ago too.
And have a function call TO and FROM that internal space
When  linear, the chromacities don't change across color spaces correct? It's just a wider or not gamut (CIE triangle).

3. The second degree of difficulty is what is the Working space for an effect?:  a clip source, the output clip... 
We might need a separate property or use probably best, the suite mechanism which also provide a way to handshake this is supported, and which would include TO and FROM
J-P was suggesting there is some sort of way to use some sort of reference mapping for that - e.g. pass the clip the color parameter is tied to.

4. What I am saying might imply that when we setValue we call TO function to pass the correct values if we performed analysis on some working space.
And when we get param value we can call FROM to be in proper current working space.

[...]

PR #244, opened 2026-06-18
Add contrib/colour header-only colour-conversion library (#158)

garyo:
Part of the Colour Managed Color Parameter work (#158). Companion to #245 (the `kOfxParamPropColourManagement` spec change).

## What

Adds `contrib/colour/ofxColourConvert.h`: a small, dependency-free, header-only C++17 library that converts RGB triplets between the OFX native colourspaces and ACES2065-1, for plug-in and host authors who want exact conversions without taking on a full OpenColorIO dependency. It is **not** a replacement for OCIO — display rendering and the display-referred / ADX / "basic" spaces are out of scope.

This is the conversion helper referenced by #245: a plug-in that declares a parameter `kOfxParamColourManagementManaged` receives its values in ACES2065-1, and can use this library to convert them to or from whatever working space it needs.

Every supported space reduces to `code value --(transfer function)--> linear RGB --(3x3 matrix)--> AP0` and back; gamut matrices are derived at compile time (NPM + von Kries adaptation) to match the OCIO ACES config the OFX native colourspaces are based on.

## Layout & integration

- `contrib/colour/ofxColourConvert.h` — the library, with a README.
- `contrib/colour/tests/` — validation suite comparing every supported colourspace against OpenColorIO in both directions, plus round-trip and name-lookup checks.
- **CMake/Conan package only the header**: an `INTERFACE` target `OpenFX::ColourConvert` that installs it, plus a `ColourConvert` Conan component. No test dependencies leak into the core package.
- **CI release tarball** ships the header under `openfx/contrib/colour/`.
- **Tests are pcons-driven** (pcons fetches OpenColorIO + GoogleTest via Conan), gated behind `OFX_BUILD_COLOUR_TESTS` (default OFF). When on, a CTest fixture delegates to `uvx pcons`, so `ctest` runs them with zero install and no impact on the main build.

## Validation

- Header compiles clean under `-Wall -Wextra`; all 39 colourspaces round-trip to 1e-9.
- Full Conan→CMake→CTest path: **77/77 gtest cases pass, 0 skipped** (every space compared against OCIO in both directions).
- `OpenFX::ColourConvert` links from a consumer; `cmake --install` and the CI tarball both place the header.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/contrib/colour/ofxColourConvert.h
+++ b/contrib/colour/ofxColourConvert.h
@@ -1,0 +1,697 @@
+#ifndef _ofxColourConvert_h_
+#define _ofxColourConvert_h_
+
+// Copyright OpenFX and contributors to the OpenFX project.
+// SPDX-License-Identifier: BSD-3-Clause
+
+/** @file ofxColourConvert.h
+
+A small, dependency-free, header-only C++17 utility for converting RGB
+triplets between the OFX "native" colourspaces (see ofxColour.h and the
+config header ofx-native-v1.5_aces-v1.3_ocio-v2.3.h) and the ACES2065-1
+(AP0) reference colourspace.
+
+This is a convenience for plug-in and host authors who want simple
+colour conversions without a full OCIO dependency. It is NOT a
+replacement for OCIO: it does not implement display rendering (the ACES
+Output Transform / RRT+ODT), so the display-referred colourspaces
+(the @c *_display spaces), the ADX film-density encodings, and the
+abstract "basic" @c ofx_* spaces are intentionally out of scope. What is
+implemented are the scene-referred working spaces and the camera/texture
+encodings, each of which reduces exactly to:
+
+    code value --(transfer function)--> linear RGB --(3x3 matrix)--> AP0
+
+and back. Every conversion here is colorimetrically exact and round-trips
+to ACES2065-1.
+
+@par Derivation
+The gamut matrices are derived at compile time from the published xy
+chromaticities of each colourspace using the Normalized Primaries Matrix
+plus a von Kries chromatic adaptation to the ACES white point, exactly as
+done by OpenColorIO's built-in transforms. The adaptation method (Bradford
+or CAT02) and the transfer-function constants match the OCIO ACES config
+that the OFX native colourspaces are based on, so results agree with OCIO
+to within floating-point precision.
+
+@par Usage
+@code
+    using namespace ofx::colour;
+    RGB c = { 0.2, 0.5, 0.8 };                       // an S-Log3/S-Gamut3 pixel
+    RGB aces = toACES2065_1(c, Colourspace::slog3_sgamut3);
+    RGB out  = fromACES2065_1(aces, Colourspace::lin_rec709_srgb);
+    // or directly:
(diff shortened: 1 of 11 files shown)


garyo:
I should note that most of this code was co-written with Claude Opus 4.8 and reviewed by me.

contrib/colour/ofxColourConvert.h
```diff
@@ -0,0 +1,697 @@
...
+    using namespace ofx::colour;
+    RGB c = { 0.2, 0.5, 0.8 };                       // an S-Log3/S-Gamut3 pixel
+    RGB aces = toACES2065_1(c, Colourspace::slog3_sgamut3);
+    RGB out  = fromACES2065_1(aces, Colourspace::lin_rec709_srgb);
+    // or directly:
+    RGB out2 = convert(c, Colourspace::slog3_sgamut3, Colourspace::lin_rec709_srgb);
+@endcode
+*/
+
+#if !defined(__cplusplus) || __cplusplus < 201703L
```
MikkiWells, line 48: The __cplusplus is only properly defined in Visual Studio if the /Zc:__cplusplus compiler option is provided.
If the option is not provided then the value of __cplusplus is 199711L. i.e. incorrect.
See https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170
To solve this you should also test for the value of _MSVC_LANG. E.g:
#if !defined(__cplusplus) || __cplusplus < 201703L || !defined(_MSVC_LANG) || _MSVC_LANG < 201703L

contrib/colour/ofxColourConvert.h
```diff
@@ -0,0 +1,697 @@
+#ifndef _ofxColourConvert_h_
+#define _ofxColourConvert_h_
```
MikkiWells, line 2: Why not use #pragma once ?

contrib/colour/ofxColourConvert.h
```diff
@@ -0,0 +1,697 @@
...
+}
+
+/** @brief Inverse of a 3x3 matrix (assumes the matrix is non-singular). */
+constexpr Matrix33 inverse(const Matrix33 &a)
+{
+    const double det =
+          a.m[0][0] * (a.m[1][1] * a.m[2][2] - a.m[1][2] * a.m[2][1])
+        - a.m[0][1] * (a.m[1][0] * a.m[2][2] - a.m[1][2] * a.m[2][0])
+        + a.m[0][2] * (a.m[1][0] * a.m[2][1] - a.m[1][1] * a.m[2][0]);
+    const double inv = 1.0 / det;
```
MikkiWells, line 99: Should there be a check for there being no defined inverse?
What if det is zero or very close to zero?

MikkiWells (changes requested):

contrib/colour/ofxColourConvert.h
garyo: I couldn't believe this was actually true, but you're 100% right. (I've now updated my `pcons` build tool to always set that flag by default.) I'll make all the suggested changes including this one.

contrib/colour/ofxColourConvert.h
garyo: done!

PR #258, opened 2026-07-22
Add signed source releases via sigstore

### LINKED ISSUE #249: Support for signed source releases
Several ASWF projects have implemented signed source releases. For instance in OpenEXR:

https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/release-sign.yml

This has a few advantages:

- immutable source release, even if the release tag gets modified after the fact
- visible SHA-256 checksum for the tarball
- ability to programmatically verify authenticity of the tarball

Unfortunately there doesn't seem to be a way to prevent the automatically generated links for unsigned .tar.gz and .zip archives from showing up.

OpenFX is already signing its binaries, so it should be straightforward to also sign the source releases.

Image

garyo:
Create an immutable source tarball at each release tag, sign it with sigstore, and attach the tarball, .sigstore.json bundle, and a .sha256 checksum to the GitHub release — following the pattern OpenEXR and other ASWF projects use. Also runnable via workflow_dispatch to sign an existing release retroactively. Uses the same pinned sigstore action the binary-signing steps in build.yml already use.

Fixes #249.

### DIFF
--- a/.github/workflows/release-sign.yml
+++ b/.github/workflows/release-sign.yml
@@ -1,0 +1,79 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright OpenFX and contributors to the OpenFX project.
+
+#
+# Create a signed, immutable source archive for each release, following
+# the pattern used by other ASWF projects (e.g. OpenEXR). See issue #249.
+#
+# This creates a .tar.gz of the complete OpenFX source tree at the given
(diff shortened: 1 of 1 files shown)


jfpanisset (approved):
Looks good, I'll go ahead and update the conanfile.yml for my OpenFX recipe to get. the newly signed releases. Thanks for doing this!

PR #262, opened 2026-08-05
Add OFX 1.6 host and parameter layout properties

fxtech:
Adds the Nuke OFX extension properties that have no current analogue:

- kOfxParamPropLayoutHint
- kOfxParamPropLayoutPadWidth
- kOfxPropHostProductTitle
- kOfxPropHostMajorVersion
- kOfxPropHostMinorVersion
- kOfxPropHostBuildVersion
- kOfxParamPropGroupIsTab

Notes:

- kOfxPropOverlayColour already had an existing equivalent: kOfxInteractPropSuggestedColour
- kOfxParamPropHidden already had an existing equivalent: kOfxParamPropSecret

This PR implements the original PR #41


revisionfx:
There is already a multi-dimensional property for host version - as well as host product name
Host Version Propertiesk
OfxPropVersion: Integer array property containing numeric version dimensions (major, minor, micro, etc.).
kOfxPropVersionLabel: String property providing a user-readable version string.kOfxPropName / kOfxPropLabel: Strings to identify the unique name or visible label of the host application.
kOfxPropAPIVersion: Integer array property for the OpenFX API version implemented by the host

fxtech:
Well-observed @revisionfx !
Sounds like we already have everything in place except maybe BuildVersion and the parameter properties - and those are kind of Nuke-specific. I could see *maybe* supporting one or two of them but as a general property - I don't know.

Thoughts/comments?

revisionfx:
You can do build number using 4D, already a host at least doing that: e.g. 2026.5.0.34

fxtech:
> You can do build number using 4D, already a host at least doing that: e.g. 2026.5.0.34

You're right! I was thinking of commit hash.
That leaves us with only the parameter layout hints.

garyo:
OK, that's great. The layout hints are: `OfxParamPropGroupIsTab` (seems like a nice idea, but there are some semantics to work out here: successive groups with this prop become a tab group I guess?), and `OfxParamPropLayoutHint`:

For LayoutHint (an enum, right?) the values in the original header are:
```
//   lay out as normal
#define kOfxParamPropLayoutHintNormal 0

//   put a divider after parameter
#define kOfxParamPropLayoutHintDivider 1

//   have the next parameter start on the same line as this
#define kOfxParamPropLayoutHintNoNewLine 2
```

I can definitely imagine wanting to use both of these in some cases. Then `OfxParamPropLayoutPadWidth` specifies the plugin's desired horizontal space between this one and the next in the NoNewLine case above — that padding width I'm less sure of, because it seems pretty host- and screen-size-specific (never mind HiDPI). 

So can we get agreement from any host to implement these two?

revisionfx:
For kOfxParamPropGroupIsTab - would be what an app like Fusion needs, we have Page parameters. But tab is a distinct UI thing. For example in Fusion native we use Tab thing equivalent.

PR #263, opened 2026-09-01
Fix incorrect metadata in documentation comments

umajho:
Hello. While working on a project where I generate code based on the new C++ bindings (`openfx-cpp/`), I found that `OfxImageEffectActionIsIdentity.outArgs` was missing in `ofxPropsBySet.h`. This turned out to be because the `@actiondef` for `kOfxImageEffectActionIsIdentity` was missing the `outArgs` metadata (`OfxPropName` and `OfxPropTime`), even though they are listed in the action's documentation.

In this PR:
- Metadata for the newly added `OfxImageEffectPropBehaviourWhenUnlicensed` are added (e49897ee0735c383ee92093ec06fa4a2c6bd0ca7)
- The two missing `outArgs` entries for `OfxImageEffectActionIsIdentity` mentioned above are added (3b09df48fee6bfe778605730b46105c46b012693)
- I asked GitHub Copilot (Kimi K3 Max) to review the metadata in the C headers in `include/`, and I fixed some issues it identified: (3a4eaca233a7cf26087241495abc801b7ef06878)
   - `OfxImageEffectPropFieldToRender`, `OfxImageEffectPropRenderWindow` and `OfxImageEffectPropRenderScale` are added to `OfxImageEffectActionRender`'s `inArgs`.
   - `OfxPropName` and `OfxPropTime` are added to `OfxCustomParamInterpFuncV1`'s `inArgs`.
   - `OfxImageEffectPropFrameRange_` in `OfxImageEffectActionGetFramesNeeded`'s `outArgs` is currently unrepresentable, therefore it is commented out following the convention.
   - `OfxImageEffectPropInteractiveRenderStatus` in `OfxImageEffectActionBeginSequenceRender` and `OfxImageEffectActionEndSequenceRender`'s `inArgs` were duplicated, and are deduplicated.
- Some more metadata update: (cd0745d4c3cbe39647399693505063c07cb84994)
   - `OfxParamPropInterpolationTime` is removed from `OfxCustomParamInterpFuncV1`'s `outArgs`.
   - In `OfxImageEffectActionGetRegionsOfInterest`'s `outArgs`, `OfxImageEffectClipPropRoI_` is renamed to `OfxImageClipPropRoI_`.
   - `@actiondef` is added to `OfxActionDialog`, `OfxActionOpenGLContextAttached` and `(k)OfxActionOpenGLContextDetached`.
- Three missing actions' `doxygendefine::` are added to `ofxImageEffectActions.rst` (b3b41cf2853365e40b2f2a32f7157ea1171f33ac)

I ran `uv run scripts/gen-props.py` and `uv run scripts/gen-props-doc.py` before submitting this PR. These should be enough for the changes in this PR.


include/ofxImageEffect.h
```diff
@@ -308,7 +308,7 @@ These are the list of actions passed to an image effect plugin's main function.
       - OfxPropTime
       - OfxImageEffectPropThumbnailRender
     outArgs:
-      - OfxImageEffectPropFrameRange
+    # - OfxImageEffectPropFrameRange_ # with clip name
```
umajho, line 311: In the same way as for `OfxImageEffectClipPropRoI_ ` in `kOfxImageEffectActionGetRegionsOfInterest`:
https://github.com/umajho/openfx/blob/e82be712c7f7bed3ba3581b42902647ce1c6fe90/include/ofxImageEffect.h#L204

include/ofxParam.h
umajho, line 934: Is`OfxParamPropInterpolationTime` here actually one of the `outArgs` for `OfxCustomParamInterpFuncV1`? The documentation doesn't list it as an `outArg`, and I can't figure out what it would mean as an `outArg`. (I'm new to OpenFX, so this might just be a gap in my understanding.)

include/ofxImageEffect.h
umajho, line 202: I think the correct name for this might be `OfxImageClipPropRoI_` (12 results in the repo) instead of `OfxImageEffectClipPropRoI_` (only used here).

include/ofxParam.h
umajho, line 1394: I think this is a typo (`s` is missing from `Params`):

```suggestion
/** @propset ParamsDouble1D
```

umajho:
(Oh… I misunderstood the purpose  of `@propsetdef`. Sorry. Please ignore this.)

include/ofxParam.h
garyo: You're right, that shouldn't be listed as an outArg.

include/ofxImageEffect.h
garyo: Yes, you're right here as well. Thanks!

garyo:
> Should I also add `@actiondef` metadata for `OfxActionDialog`, `OfxActionOpenGLContextAttached` and `OfxActionOpenGLContextDetached`?
> 
> The documentation page [Actions Passed to An Image Effect](https://openfx.readthedocs.io/en/latest/Reference/ofxImageEffectActions.html) doesn't mention these actions. Is there a reason for this? For example, did I misunderstand their purpose: are they not actions that are passed to`mainEntry`?

Yes! Good catch. That's historical drift. All three actions have no props, no inArgs/outArgs. Adding those would be great here.

One note to remember: the detached action's string value is "kOfxActionOpenGLContextDetached", with a stray leading "k". That's a long-standing bug that is frozen for compatibility and we can't change it.

garyo:
> The metadata for the texture property set that appears in functions like `OfxImageEffectOpenGLRenderSuiteV1::clipLoadTexture` are missing. Should I also add them here?

You retracted for the right reason but you spotted a real issue: missing Texture propset for the propset returned by load-texture call, which has lots of useful properties. So your instinct was right even though @propsetdef was the wrong tag: that one defines reusable fragments that other sets include. Property sets themselves are declared with @propset (see the Image set at the end of ofxImageEffect.h).

There is indeed no @propset for the texture handle returned by clipLoadTexture, so the bindings don't generate the accessors for it. I suggest adding this as a separate PR, if you're interested, so we can review it independently. Basically: a "Texture" @propset in ofxGPURender.h (write: host) listing the 12 properties documented on clipLoadTexture,
plus OfxPropType, and a note on the two OpenGL texture properties that they also live on the texture handle.

garyo:
Note re: rebase conflicts, when you get to it: the conflicts will be in the 3 generated files, so to make it easy, just take either side temporarily, then rerun the generators and commit the result.

umajho:
Thank you for the advices! I have updated the PR accordingly.

I think I can submit another PR for the missing Texture propset later.

And sorry for the inconvenience caused by using `main` as my local branch's name. I didn't realize it until I updated the branch…

include/ofxDialog.h
```diff
@@ -39,6 +39,10 @@ actions needed. (Like lowering its priority etc..)
 	   Plugin should return from this action when all Dialog interactions are done.
 	   At that point the host will continue again.
 	   The host will not send any other messages asynchronous to this one.
+ 
+    @actiondef
+    inArgs:
```
umajho, line 44: Maybe in the future there should be a way to express that `inArgs` here is not a property set but `void *`…?

garyo: Good point. One could argue that user data there should have been a property on inArgs rather than a void*, but that can't be changed now. I think empty inArgs: is OK for now but if you want to file an issue on this we can take a look.

Also a separate PR for Texture propset would be appreciated!

umajho:
Oh, I found a few issues while trying to use the updated `ofxPropsBySet.h`. I will leave my findings as review comments below.

I'm not familiar with how the code generator produces `ofxPropsBySet.h`, but I can try to open a PR to fix these issues if you'd like.

openfx-cpp/include/openfx/ofxPropsBySet.h
```diff
@@ -706,6 +708,7 @@ static inline const std::array<const char *, 33> actions {
   "OfxInteractActionPenDown",
   "OfxInteractActionPenMotion",
   "OfxInteractActionPenUp",
+  kOfxActionOpenGLContextDetached,
```
umajho, line 711: Since `ofxPropsBySet.h‎` includes `ofxGPURender.h`, the value here is technically correct, however, shouldn't this instead be a string literal (`"kOfxActionOpenGLContextDetached"`)?

openfx-cpp/include/openfx/ofxPropsBySet.h
```diff
@@ -935,10 +943,12 @@ static_assert(std::string_view("OfxActionBeginInstanceEdit") == std::string_view
 static_assert(std::string_view("OfxActionCreateInstance") == std::string_view(kOfxActionCreateInstance));
 static_assert(std::string_view("OfxActionDescribe") == std::string_view(kOfxActionDescribe));
 static_assert(std::string_view("OfxActionDestroyInstance") == std::string_view(kOfxActionDestroyInstance));
+static_assert(std::string_view("OfxActionDialog") == std::string_view(kOfxActionDialog));
```
umajho, line 946: `ofxDialog.h`, where kOfxActionDialog is defined, is not included by `ofxPropsBySet.h`.

openfx-cpp/include/openfx/ofxPropsBySet.h
```diff
@@ -935,10 +943,12 @@ static_assert(std::string_view("OfxActionBeginInstanceEdit") == std::string_view
...
 static_assert(std::string_view("OfxActionEndInstanceChanged") == std::string_view(kOfxActionEndInstanceChanged));
 static_assert(std::string_view("OfxActionEndInstanceEdit") == std::string_view(kOfxActionEndInstanceEdit));
 static_assert(std::string_view("OfxActionInstanceChanged") == std::string_view(kOfxActionInstanceChanged));
 static_assert(std::string_view("OfxActionLoad") == std::string_view(kOfxActionLoad));
+static_assert(std::string_view("OfxActionOpenGLContextAttached") == std::string_view(kOfxActionOpenGLContextAttached));
```
umajho, line 951: missing:
`static_assert(std::string_view("kOfxActionOpenGLContextDetached") == std::string_view(kOfxActionOpenGLContextDetached));`

